PR #41, opened 2022-11-23
Static reflection

MartinMystikJonas:


src/Template/Variable.php
```diff
@@ -24,6 +24,11 @@ public function getName(): string
         return $this->name;
     }
 
+    public function getType(): Type
```
lulco, line 27: do we need this? it is not used

lulco:
please fix sniffer

src/Template/Variable.php
MartinMystikJonas: shit it was commited by mistake. I need it in that variable type merging I was working on and I commited it in this PR by mistake. Should I remove it or will we keep it here?

MartinMystikJonas:
@lulco CS fixed

src/Template/Variable.php
lulco: If you will need it later, we can merge

PR #45, opened 2022-11-24
Configurable tmpDir and context hash in compiled templates

MartinMystikJonas:


src/Template/Component.php
```diff
@@ -28,4 +29,9 @@ public function getTypeAsString(): string
     {
         return $this->type->describe(VerbosityLevel::typeOnly());
     }
+
+    public function jsonSerialize(): mixed
+    {
+        return [$this->name, $this->getTypeAsString()];
```
lulco: I would prefer to use also some keys in arrays 'name' => $this->name, 'type' => $this->getTypeAsString()

src/Template/Template.php
```diff
@@ -45,4 +47,9 @@ public function getComponents(): array
     {
         return $this->components;
     }
+
+    public function jsonSerialize(): mixed
+    {
+        return [$this->path, $this->variables, $this->components];
```
lulco: keys

src/Compiler/LatteToPhpCompiler.php
```diff
@@ -56,6 +60,37 @@ public function compile(Scope $scope, string $templateContent, array $variables,
...
     }
 
+    /**
+     * @param Variable[] $variables
+     * @param Component[] $components
+     */
+    public function compileFile(Scope $scope, string $templatePath, array $variables, array $components): string
+    {
+        $templateContent = file_get_contents($templatePath) ?: '';
+        $phpContent = $this->compile($scope, $templateContent, $variables, $components);
```
lulco, line 70: compile should be private then

src/Rule/LatteTemplatesRule.php
```diff
@@ -43,7 +43,7 @@ public function __construct(
         FileAnalyserFactory $fileAnalyserFactory,
         RuleRegistry $rulesRegistry,
         CollectorsRegistry $collectorsRegistry,
-        ErrorBuilder $errorBuilder
+        ErrorBuilder $errorBuilder,
```
lulco: we are on php 7.4

src/Template/Variable.php
```diff
@@ -33,4 +34,9 @@ public function getTypeAsString(): string
     {
         return $this->type->describe(VerbosityLevel::typeOnly());
     }
+
+    public function jsonSerialize(): mixed
+    {
+        return [$this->name, $this->getTypeAsString()];
```
lulco: keys

lulco:
also check the pipeline

src/Compiler/LatteToPhpCompiler.php
MartinMystikJonas: It is used in tests, I would keep it public

MartinMystikJonas:
@lulco Pipeline fixed

src/Compiler/LatteToPhpCompiler.php
lulco: Then we should create some other service just to store files, compiler just creates php content and that new service will just save it to file.

MartinMystikJonas: It seems as unnecessary complication for me.

PR #46, opened 2022-11-24
Collectors

### LINKED ISSUE #30: Variables and paths are not found if template is assigned to local variable
```
$template = $this->template;
$template->foo = 'bar';
$template->setFile(__DIR__ . '/default.latte');
```
not working

### LINKED ISSUE #39: Recursive finding of components
Like with variables, we need to go deeper

```php
final class ComponentsPresenter extends ParentPresenter
{
    public function actionDefault(): void
    {
        parent::actionDefault();
        // do something
    }
}

abstract class ParentPresenter extends Presenter
{
    protected function startup()
    {
        parent::startup();
        $this->addComponent(new Form(), 'startupForm');
    }

    public function actionDefault(): void
    {
        $this->addComponent(new Form(), 'onlyParentDefaultForm');
    }
}
```

### LINKED ISSUE #43: Template variable assignments when template is passed as parameter
```php
public function render() {
  $this->setTemplateData($this->template);
}

public function setTemplateData($template) {
  $template->something = "something";
}

```

lulco:
@MartinMystikJonas check this.  Everything what needs scope is collected in collectors - variables, method calls, components etc. Then in resolvers we can use these collected items out of context (see https://phpstan.org/developing-extensions/collectors "Because these custom CollectedDataNode rules are executed out of context,...").

Resolves https://github.com/efabrica-team/phpstan-latte/issues/30
Resolves https://github.com/efabrica-team/phpstan-latte/issues/39
Resolves https://github.com/efabrica-team/phpstan-latte/issues/43 - if type is set to called method


lulco:
Hm it's not working on my real-world application. Only with —debug :( have to check it

lulco:
Probably some file cache issues.

PR #50, opened 2022-11-26
Combine collected variable types

MartinMystikJonas:


src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
```diff
@@ -5,12 +5,14 @@
 namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor;
 
 use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\ActualClassNodeVisitorBehavior;
+use Efabrica\PHPStanLatte\Template\Variable;
```
lulco, line 8: Duplicity, one row under is the same use but aliased. Delete one of them (probably aliasing is no need anymore)

lulco: I merged it now, cleanup can be done later. Now I have to go to trip with family :)

PR #54, opened 2022-11-26
Fixed caching problems when storing objects in collector results

MartinMystikJonas:


MartinMystikJonas:
Alternative solution instead of $forgotX. Add toArray() method (same as jsonSerialioze) and return only this array from Collector? Probably cleaner solution I think.

MartinMystikJonas:
I will update this PR

lulco:
Great, thank you. One more thing: I think, now we don’t need jsonSerialize methods in Collector Value Objects because they are not serialized now (need to check if it is true). I've added them before because of some caching issues

MartinMystikJonas:
Ok i removed them

lulco:
Thank you very much!

PR #55, opened 2022-11-26
Recursion fix

MartinMystikJonas:
Includes caching fix https://github.com/efabrica-team/phpstan-latte/pull/54


lulco:
You can rebase to clean diff. Shouldn't be arrays passed by reference?

MartinMystikJonas:
No need for that but it would lower memory use so I will change it

lulco:
Great work! You don’t need me anymore :D

PR #64, opened 2022-11-28
Resolvers reworked

MartinMystikJonas:
Some cahnges to resolvers to enable easier writing of custom resolvers

1) Resolvers can collect any node not just class (some resolvers could rely on single method or call) - required cahbge from CollectedClass to CollectedNode with array or params
2) Single node can be resolved by multiple resolvers (custom resolvers can be addition to default resolvers on same node)
3) Actual class is optional - because there could be cases when latte is compiled without control/presenter (in our case it could be for example e-mail content)
4) Extracted common code of class-based resolvers to abstract class
5) Some structural improvements


MartinMystikJonas:
Let me know what do oyu think. I would like to finish this before I start writing custom resolvers needed to properly check our projects.

lulco:
Seems resonable to me. I'll check the code later, but I think the description is clear and correct. Can you draft some readme for custom resolvers, too?

MartinMystikJonas:
I am planning to. Bur I think rhere woukd be some more changes so I will do it when I would have some working custom resolver.

MartinMystikJonas:
rebased onto main

MartinMystikJonas:
Should be ready to merge

PR #68, opened 2022-11-28
Define types of items in $this->global in compiled template

MartinMystikJonas:


tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -27,6 +27,26 @@ public function testVariables(): void
                 5,
                 'default.latte',
             ],
+            [
```
lulco, line 30: I would like to not report issues which user/developer can't fix from latte. These are some internal nette/latte errors right? We should fix them by modifying compiled template or ignore them.

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -143,6 +163,11 @@ public function testComponents(): void
     public function testFilters(): void
     {
         $this->analyse([__DIR__ . '/Fixtures/FiltersPresenter.php', __DIR__ . '/Fixtures/ParentPresenter.php'], [
+            [
+                'Trying to invoke mixed but it\'s not a callable.',
```
lulco, line 167: this is covered by message Undefined latte filter, imho we can ignore it

src/VariableCollector/DefaultTemplateVariables.php
```diff
@@ -21,6 +21,7 @@ public function collect(): array
         $variables[] = new Variable('baseUrl', new StringType());
         $variables[] = new Variable('basePath', new StringType());
         $variables[] = new Variable('ʟ_fi', new ObjectType('Latte\Runtime\FilterInfo'));
+        $variables[] = new Variable('ʟ_tag', new ArrayType(new MixedType(), new MixedType()));
```
lulco: don't we know the exact or more strict type of this variable?

lulco (commented):
There are few errors in unit tests, check them please

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
MartinMystikJonas: In this case yes but same error can occur in different context caused by user code error

src/VariableCollector/DefaultTemplateVariables.php
MartinMystikJonas: I think it is string[] but not quite sure without closet examination of latte internals

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
MartinMystikJonas: It is not latte internal but secondary errors caused by use of unknown (mixed) variable. It is imho expected to occur on level 9. Same errors would occur in php code when foreach iterates on mixed and you try access properties of mixed.

MartinMystikJonas:
I will check tests errors - weirs that it did not occured in local. Maybe different version of some dependency.

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
lulco: ok then

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
MartinMystikJonas: For example when user passes callable in variable and tries to invoke it in template then same error will occur. So we cannot ignore it. Only option would be conditional ignore I mwntioned - ignore this error only if undefined filter error is present on same location

src/VariableCollector/DefaultTemplateVariables.php
MartinMystikJonas: It seems to be string[] I will narrow it down.

lulco (approved):

PR #70, opened 2022-11-28
Check included latte files

### LINKED ISSUE #22: Check included latte files

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/22


src/Rule/LatteTemplatesRule.php
```diff
@@ -71,28 +81,49 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
...
+            } catch (Throwable $e) {
+                $errors = array_merge($errors, $this->errorBuilder->buildErrors([new Error($e->getMessage(), $scope->getFile())], $templatePath, $scope));
+                continue;
+            }
+
+            $fileAnalyserResult = $this->fileAnalyserFactory->create()->analyseFile(
+                $compileFilePath,
+                [],
+                $this->rulesRegistry,
+                new CollectorsRegistry([$this->includePathCollector]),
```
MartinMystikJonas, line 111: Why is new Collectors registry created instead of using existing like before?

src/Rule/LatteTemplatesRule.php
```diff
@@ -71,28 +81,49 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
...
+
+            $dir = dirname($templatePath);
+
+            $collectedDataList = $fileAnalyserResult->getCollectedData();
+            $includeTemplates = [];
+            foreach ($collectedDataList as $collectedData) {
+                /** @phpstan-var CollectedIncludePathArray $data */
+                $data = $collectedData->getData();
+                $collectedIncludedPath = CollectedIncludePath::fromArray($data, $this->typeStringResolver);
+                $includeTemplates[] = new Template($dir . '/' . $collectedIncludedPath->getPath(), $template->getActualClass(), $collectedIncludedPath->getVariables() + $template->getVariables(), $template->getComponents());
```
MartinMystikJonas: I would either create Template::fromCollectedIncludeTemplate() or CollectedIncludeTempalte::getTemplate() and movet his logic there

src/Collector/ValueObject/CollectedIncludePath.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * @phpstan-param CollectedIncludePathArray $item
+     */
+    public static function fromArray(array $item, TypeStringResolver $typeStringResolver): self
+    {
+        $variables = [];
+        foreach ($item['variables'] as $variable) {
+            $variables[] = new Variable($variable['variableName'], $typeStringResolver->resolve($variable['variableType']));
```
MartinMystikJonas: Create Variable fromArray and toArray as for value objects to keep object<->array transformations in value object itself

src/Rule/LatteTemplatesRule.php
```diff
@@ -71,28 +81,49 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
...
+            }
+            $this->analyseTemplates($includeTemplates, $scope, $errors);
```
MartinMystikJonas, line 127: Is there danger of infinite loop? Could two templates include each other conditionally?

src/Rule/LatteTemplatesRule.php
lulco: because we don't need all of them, just include path collector here. It is some micro optimization - there are no method calls, variables, etc analysed on compiled template

src/Rule/LatteTemplatesRule.php
lulco: well, like always, there can be infinite loop. Then there is question when to stop it in static analysis. We can't stop analysing of included template just because it was already analysed (in different context)

MartinMystikJonas: I would stop search for included files in files already searched for included files. So will will analyse A called from PHP, found call of B, analyse B called from A, found call of A, analyse A called from B, but we will not againg analyse A to find call of B from A again

MartinMystikJonas: Also reset list for teach resolved node

lulco: have to think about it, I'll check some projects if there is recursion possible in nette / latte itself

PR #74, opened 2022-11-29
Include recursion fix

MartinMystikJonas:


tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -318,4 +318,9 @@ public function testLinks(): void
             ],
         ]);
     }
+
+    public function testRecursion(): void
+    {
+        $this->analyse([__DIR__ . '/Fixtures/RecursionPresenter.php', __DIR__ . '/Fixtures/ParentPresenter.php'], []);
```
lulco, line 324: You can dump type of counter to prove that it is 10, 9, 8, 7?

lulco (approved):

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
MartinMystikJonas: It dumps just int because it it is passed variable with compile time unknown value

lulco: Than dump at least int and we will check that it is 7 times (or how many?) dumped

MartinMystikJonas: 4 times 1st for original call and then 3 times depth

Issue #69, opened 2022-11-28
Detect parameters passed in parameter #2 or render() method


MartinMystikJonas:
Shoulkd also include two layer resolution of template path - collect calls to render and only if render has no parameter or fist parameter is null use path collected from setFile (parameter of render call takes precedence, and passing null to render is fine only when setFile is called instead)

PR #81, opened 2022-11-30
Show errors when template resolving fails

MartinMystikJonas:
- LatteTemplateResolver return LatteTemplateResolver that can contain errors
- TemplatePathFinder can return tempalte with null path to signal failure to evaluate expression

- Show error when template path is expression that cannot be resolved automatically
- Show error when no render/setFile call is found for control render
- Show error when no template is found for presenter action

NEW
- fixed finding variables/componets also in parents
- fixed finding template paths also in called methods
- fixed error when render method calls another render method
- fixed error when same tempalte with same variables and components is resolved multiple times in different contexts

TODO:
- [x] expresisons containing variables are resolved as if variables are empty (maybe support replacing of variable with wildcard and look for tempaltes matching pattern?)
- [] better context info about location of unresolved expression
- [x] would it be possible to interpret expression when it is assigned to variable before passing to render/setFile? (moved to #92)
- [x] ignore presenter action that terminates early (redirect, sendResponse - detect call of method with never return type?)
- [x] two layer resolution of template path - collect calls to render and only if render has no parameter or fist parameter is null use path collected from setFile (parameter of render call takes precedence, and passing null to render is fine only when setFile is called instead) (moved to #69)
- [x] NetteApplicationUIControl also checks render* methods in presenters - is it by mistake or intentional? If by mistake then add condition to exclude it, if intentional the do not report that tempalte cannot be resolved because it is covered by NetteApplicationUIPresenter

### DIFF
--- a/src/Collector/Finder/ComponentFinder.php
+++ b/src/Collector/Finder/ComponentFinder.php
@@ -7,6 +7,7 @@ namespace Efabrica\PHPStanLatte\Collector\Finder;
 use Efabrica\PHPStanLatte\Collector\ComponentCollector;
 use Efabrica\PHPStanLatte\Collector\ValueObject\CollectedComponent;
 use Efabrica\PHPStanLatte\Template\Component;
+use PHPStan\BetterReflection\BetterReflection;
 use PHPStan\BetterReflection\Reflection\ReflectionMethod;
 use PHPStan\Node\CollectedDataNode;
 use PHPStan\PhpDoc\TypeStringResolver;
@@ -19,7 +20,7 @@ final class ComponentFinder
     /**
      * @var array<string, array<string, Component[]>>
      */
-    private array $collectedComponents;
+    private array $collectedComponents = [];
 
     private MethodCallFinder $methodCallFinder;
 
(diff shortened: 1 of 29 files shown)


MartinMystikJonas:
Do you approve this approach? If yes I will rebase nad fix tests

lulco:
I like that we could have error when no template is found, just don't like that there is `|` in return type :) Template and Error are completely different interfaces. We have to check it always. Is there another way how to achieve our target?

MartinMystikJonas:
Well another option is to have ResolverResult object that contains Templates[] and Errors[]

lulco:
> Well another option is to have ResolverResult object that contains Templates[] and Errors[]

and we will have some getters for both? It's better isn't it?

MartinMystikJonas:
OK I will change it it would be also robust for future extension of resolvers

MartinMystikJonas:
NEW
- fixed finding variables/componets also in parents
- fixed finding template paths also in called methods
- fixed error when render method calls another render method
- fixed error when same tempalte with same variables and components is resolved multiple times in different contexts

lulco (approved):

PR #82, opened 2022-11-30
Resolvers refactoring

MartinMystikJonas:


src/LatteTemplateResolver/AbstractClassMethodTemplateResolver.php
```diff
@@ -0,0 +1,32 @@
...
+
+abstract class AbstractClassMethodTemplateResolver extends AbstractClassTemplateResolver
+{
+    protected function getClassTemplates(ReflectionClass $reflectionClass, CollectedDataNode $collectedDataNode): array
+    {
+        $globalVariables = $this->getClassGlobalVariables($reflectionClass);
+        $globalComponents = $this->getClassGlobalComponents($reflectionClass);
+
+        $templates = [];
+        foreach ($this->getMethodsMatching($reflectionClass, $this->getClassMethodPattern() . 'i') as $reflectionMethod) {
```
lulco, line 19: why case insensitive?

src/LatteTemplateResolver/AbstractClassTemplateResolver.php
```diff
@@ -42,12 +44,25 @@ public function collect(Node $node, Scope $scope): ?CollectedResolvedNode
...
             if ($objectType->isInstanceOf($supportedClass)->yes()) {
-                return new CollectedResolvedNode(static::class, [self::PARAM_CLASS_NAME => $className]);
+                $isSupported = true;
+            }
+        }
+
+        $isIgnored = false;
+        foreach ($this->getIgnoredClasses() as $ignoredClass) {
+            if ($objectType->isInstanceOf($ignoredClass)->yes()) {
+                $isIgnored = true;
```
lulco: we can do break here

src/LatteTemplateResolver/AbstractClassTemplateResolver.php
```diff
@@ -42,12 +44,25 @@ public function collect(Node $node, Scope $scope): ?CollectedResolvedNode
         }
 
         $objectType = new ObjectType($className);
+
+        $isSupported = false;
         foreach ($this->getSupportedClasses() as $supportedClass) {
             if ($objectType->isInstanceOf($supportedClass)->yes()) {
-                return new CollectedResolvedNode(static::class, [self::PARAM_CLASS_NAME => $className]);
+                $isSupported = true;
```
lulco: break?

src/LatteTemplateResolver/NetteApplicationUIControl.php
```diff
@@ -4,41 +4,22 @@
...
 {
+    use NetteApplicationUIControlGlobals;
+
     public function getSupportedClasses(): array
     {
         return ['Nette\Application\UI\Control'];
     }
 
-    protected function getClassTemplates(ReflectionClass $reflectionClass, CollectedDataNode $collectedDataNode): array
+    public function getIgnoredClasses(): array
```
lulco, line 16: should be possible to set from config:
example: I have Plugins which are also Controls and I want to resolve them by my own resolver not this built-in

src/LatteTemplateResolver/NetteApplicationUIPresenter.php
```diff
@@ -74,14 +50,16 @@ private function findTemplateFilePath(ReflectionClass $reflectionClass, string $
...
-        $dir = $this->findTemplateDir($reflectionClass);
+        $dir = $this->getClassDir($reflectionClass);
         if ($dir === null) {
             return null;
         }
 
         $templateFileCandidates = [
             $dir . '/templates/' . $presenterName . '/' . $actionName . '.latte',
             $dir . '/templates/' . $presenterName . '.' . $actionName . '.latte',
+            $dir . '/../templates/' . $presenterName . '/' . $actionName . '.latte',
```
lulco: I think this is not true - it is not the same as before. You are finding dir for all actions. But in Nette\UI\Presenter there is this code:
```
$dir = dirname(static::getReflection()->getFileName());
$dir = is_dir("$dir/templates") ? $dir : dirname($dir);
$list = [
	"$dir/templates/$presenter/@$layout.latte",
	"$dir/templates/$presenter.@$layout.latte",
];
```

So if `$dir/templates` it is used for all actions

src/LatteTemplateResolver/AbstractClassMethodTemplateResolver.php
MartinMystikJonas: Because method names are case insensitive in PHP

src/LatteTemplateResolver/NetteApplicationUIControl.php
MartinMystikJonas: Agreed but for now add it as issue and we will figure it out later. Ok?

src/LatteTemplateResolver/NetteApplicationUIPresenterStandalone.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        return ['Nette\Application\UI\Presenter'];
+    }
+
+    protected function getTemplatePathPatterns(ReflectionClass $reflectionClass, string $dir): array
+    {
+        $shortClassName = $reflectionClass->getShortName();
+        $presenterName = str_replace('Presenter', '', $shortClassName);
+
+        return [
```
lulco, line 25: here should be the same logic as in non standalone presenters

src/LatteTemplateResolver/NetteApplicationUIControl.php
lulco: ok

lulco (approved):

PR #84, opened 2022-11-30
Resolve all possible paths (optionally)

### LINKED ISSUE #73: Report unchecked latte files (optionally)
Find all latte files in analysedPaths. When file is analysed remove it from list. Then report all files left on list because they were unchecked. Make it optional but it would be very useful to find blind spots of analysis caused by some non standard render calls (for example with dynamically generated template path) .

MartinMystikJonas:
Resolved #73


MartinMystikJonas:
Depends on  #82

src/Resolver/ValueResolver/ValueResolver.php
```diff
@@ -10,18 +10,50 @@
...
+    public function resolveName(Expr $expr, ?string $actualFile = null)
+    {
+        return $this->resolve($expr, $actualFile, false);
+    }
+
     /**
      * @return mixed
      */
-    public function resolve(Expr $expr, ?string $actualFile = null)
+    public function resolvePath(Expr $expr, ?string $actualFile = null)
```
lulco: I think there should be some PathResolver which will use ValueResolver

src/Rule/LatteTemplatesRule.php
```diff
@@ -131,7 +132,24 @@ private function analyseTemplates(array $templates, Scope $scope, array &$errors
                 /** @phpstan-var CollectedIncludePathArray $data */
                 $data = $collectedData->getData();
                 $collectedIncludedPath = CollectedIncludePath::fromArray($data, $this->typeStringResolver);
-                $includeTemplates[] = new Template($dir . '/' . $collectedIncludedPath->getPath(), $template->getActualClass(), array_merge($collectedIncludedPath->getVariables(), $template->getVariables()), $template->getComponents());
+                if (strpos($templatePath, '*') !== false) {
+                  /** @var string $file */
```
lulco: two spaces to the right

extension.neon
```diff
@@ -10,6 +10,7 @@ parameters:
         engineBootstrap: null
         macros: []
         extensions: []
+        resolveAllPossiblePaths: false
```
lulco, line 13: please update readme

src/Collector/Finder/TemplatePathFinder.php
```diff
@@ -28,7 +30,17 @@ public function __construct(CollectedDataNode $collectedDataNode)
             if (!isset($this->collectedTemplatePaths[$className][$methodName])) {
                 $this->collectedTemplatePaths[$className][$methodName] = [];
             }
-            $this->collectedTemplatePaths[$className][$methodName][] = $collectedTemplatePath->getTemplatePath();
+            $templatePath = $collectedTemplatePath->getTemplatePath();
+            if (strpos($templatePath, '*') !== false) {
+                $dirWithoutWildcards = (string)Strings::before((string)Strings::before($templatePath, '*'), '/', -1);
+                $pattern = substr($templatePath, strlen($dirWithoutWildcards) + 1);
+                /** @var string $file */
```
lulco, line 37: why are you adding this php doc and then cast $file to string? casting is just good enough isn't it?

src/Collector/TemplatePathCollector.php
```diff
@@ -74,8 +74,8 @@ public function processNode(Node $node, Scope $scope): ?array
         }
 
         /** @var string|null $path */
-        $path = $this->valueResolver->resolve($arg->value, $scope->getFile());
-        if ($path === null) {
+        $path = $this->valueResolver->resolvePath($arg->value, $scope->getFile());
+        if ($path === null || $path[0] !== '/') {
```
lulco, line 77: $path[0] is not set for empty string

src/Collector/Finder/TemplatePathFinder.php
MartinMystikJonas: PHPStan complained becuse it does not know type of $file

src/Collector/TemplatePathCollector.php
MartinMystikJonas: empty string cannot be returned here it is covered by resolvePath

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: I was thinking abut multiple methods in ValueResolver for different types of values

MartinMystikJonas: Problem is that Value resolver would need to mae some internal parameters public to allow change of behaviour as required by PathResolver - I think one value resolver with multiple methods is cleaner

lulco: ValueResolver just resolves the value of Expression, PathResolver knows that Expression should be path and act with it as with path. For me it is clear to do not mix these two approaches

MartinMystikJonas: Problem is that I would need to pass $allowPathWildcards into Value resolver somehow and I do not like to have it in public interface of ValueResolver

MartinMystikJonas: Other option is to have duplicated code in two resolvers. Form these three options 1) dumplicated code 2) ugly internal parameter in value resolver 3) two methods in Valueresolver i chosed number 3

MartinMystikJonas: Which would you preffer: Maybe option 2 with some better behaviour? Like parameter $unknownValuePlaceholder and PathResolver would pass *?

MartinMystikJonas: Changed

MartinMystikJonas:
Should be ready to merge. Can you check reworked ValueResolver/PathResolver if it is ok this way?

src/Resolver/ValueResolver/ValueResolver.php
```diff
@@ -35,7 +37,11 @@ public function resolve(Expr $expr, ?string $actualFile = null)
             }
 
             if ($expr instanceof Cast) {
-                return $this->resolve($expr->expr, $actualFile);
+                return $this->resolve($expr->expr, $actualFile, $unknownValuePlaceholder);
+            }
+
+            if ($expr instanceof Variable && $unknownValuePlaceholder) {
```
lulco, line 43: We can resolve variable expression in Path resolver and we don’t need pass $unknownValuePlaceholder to ValueResolver

src/Rule/LatteTemplatesRule.php
```diff
@@ -131,7 +132,24 @@ private function analyseTemplates(array $templates, Scope $scope, array &$errors
                 /** @phpstan-var CollectedIncludePathArray $data */
                 $data = $collectedData->getData();
                 $collectedIncludedPath = CollectedIncludePath::fromArray($data, $this->typeStringResolver);
-                $includeTemplates[] = new Template($dir . '/' . $collectedIncludedPath->getPath(), $template->getActualClass(), array_merge($collectedIncludedPath->getVariables(), $template->getVariables()), $template->getComponents());
+                if (strpos($templatePath, '*') !== false) {
+                    /** @var string $file */
+                    foreach (Finder::findFiles($templatePath)->from($dir) as $file) {
+                        $includeTemplates[] = new Template(
```
lulco: There are not included files or are they? Maybe this should be few rows above in main foreach

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: How would you do it wirhour copy paste entire evaluator?

src/Rule/LatteTemplatesRule.php
MartinMystikJonas: It shoukd bw removed I was testing it for inckuded files too but was not able to make it useful. This is forgotten relict

src/Resolver/ValueResolver/ValueResolver.php
lulco: I have to check it on desktop, CR on mobile really sucks

lulco: Do not copy entire evaluator, just add this condition to path resolver

```
if ($expr instanceof Variable && $this->resolveAllPossiblePaths) {
$result = '*'
} else {
// call value resolver
}
```

lulco:
I like the approach, just don't like mixing some PathResolver settings to ValueResolver. Maybe we should use * always when ValueResolver returns null... what do you think?

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: But this piece of code is inside callback passed to ConstExprEvaluator.

MartinMystikJonas: I would need to extract this callback to be able do add piece of code to it

MartinMystikJonas: and it contains recursive call that would neet to recursively call pathresolver

lulco: I know, but that callback is there just because ConstExprEvaluator can't resolve Variable Node. We can handle it before calling ValueResolver. Recursive call is only one as I saw, and it is Casting, so it can be here too.

BUT what about this suggestion https://github.com/efabrica-team/phpstan-latte/pull/84#issuecomment-1333722634?

MartinMystikJonas:
It does notmake sense with variable or control names for example to contain * there we want to know it was not resolved

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: MAybe I am missing something. Can you show me how would you rewrite it?

src/Resolver/ValueResolver/PathResolver.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $this->resolveAllPossiblePaths = $resolveAllPossiblePaths;
+        $this->valueResolver = $valueResolver;
+    }
+
+    /**
+     * @return ?string
+     * @phpstan-return ?non-empty-string
+     */
+    public function resolve(Expr $expr, ?string $actualFile = null)
```
lulco, line 25: Something like:

```
    public function resolve(Expr $expr, ?string $actualFile = null)
    {
       if ($expr instanceof Variable) {
            $result = $this->resolveAllPossiblePaths ? '*' : null;
       } elseif ($expr instanceof Cast) {
             $result = $this->resolve($expr->expr, $actualFile);
        } else {
             $result = $this->valueResolver->resolve($expr, $actualFile);
        }
        if (!is_string($result)) {
            return null;
        }
        $result = preg_replace('#\*+#', '*', $result);
        if ($result === null || $result === '' || $result[0] === '*') {
            return null;
        }
        return $result;
    }
```

MartinMystikJonas: That would fail if cast is not a top level node of evaluated expression no?

MartinMystikJonas: Or if variable is not a top level node

MartinMystikJonas: We need to use different evaluation logic for all nodes in expression tree

lulco:
OK so if we can’t find nicer solution and we really need this, you can merge this PR

MartinMystikJonas:
Ok I will rebase and merge it this way.

I did somw digging in PHPStan internals and I think we can rewrite it with better capabilities (like remember evakuated value of variable set before) using  https://phpstan.org/blog/preprocessing-ast-for-custom-rules

But it could be challenge to do it right so I will left it for future

PR #87, opened 2022-12-01
Option to report unanalysed templates

MartinMystikJonas:
I was not able to find out how to test it because rule tests forces analysedPaths to empty array

### DIFF
--- a/src/Rule/LatteTemplatesRule.php
+++ b/src/Rule/LatteTemplatesRule.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Efabrica\PHPStanLatte\Rule;
 
+use Efabrica\PHPStanLatte\Analyser\AnalysedTemplatesRegistry;
 use Efabrica\PHPStanLatte\Analyser\FileAnalyserFactory;
 use Efabrica\PHPStanLatte\Collector\Finder\ResolvedNodeFinder;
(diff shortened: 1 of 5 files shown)


MartinMystikJonas:
Tested it on real world app and seems it works like a charm

src/Rule/LatteTemplatesRule.php
```diff
@@ -84,6 +90,14 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
                 $this->analyseTemplates($templates, $scope, $errors);
             }
         }
+
+        foreach ($this->analysedTemplatesRegistry->getReportedUnanalysedTemplates() as $templatePath) {
+            $errors[] = RuleErrorBuilder::message('Latte template ' . pathinfo($templatePath, PATHINFO_BASENAME) . ' was not analysed.')
+                ->file($templatePath)
+                ->line(1)
```
lulco: maybe -1 or null or nothing?

lulco (approved):

src/Rule/LatteTemplatesRule.php
MartinMystikJonas: or skip completely it is not mandatory

lulco: yes that's what I ment with "nothing"

PR #89, opened 2022-12-01
LinkProcessor parameters processing fix and fallback to static reflection

MartinMystikJonas:
Relying on native autoload of app classes in PHPStan context can lead to false positive link errors


MartinMystikJonas:
Found some secondary errors. Need to investigate

MartinMystikJonas:
Found it. There was forgotten throw I for some reson added causing error. And when exception without message is catched it is reported by exception class

MartinMystikJonas:
Found few cases where it still fails

MartinMystikJonas:
All fixed ready to merge

lulco (approved):

PR #94, opened 2022-12-02
Improved ValueResolver - uses informations from Scope

MartinMystikJonas:


lulco (approved):

MartinMystikJonas:
ValuResolver can resolve multiple possible values if PHPStan knows them  as UnionType of ConstantScalarType. Does not cover cases when type of variable is not evaluated as constant because it used something that is not constant according to PHPStan (line __DIR__) but it is at least something and we might be able to improve it in future

PR #95, opened 2022-12-02
Forms support

### LINKED ISSUE #77: Add support for forms
- [x] detect forms and their controls
- [x] create fake classes in compiled templates
- [x] change `end($this->global->formsStack)` to `$form`
- [x] change `$form = $this->global->formsStack[] = $this->global->uiControl["form"]` to `$form = new FakeClass() from step 2`
- [x] profit

FakeClass example:
```
class CopyForm_61aa5bae0c1dd5664908f465d1d41866 extends Form
{
    /**
     * @param $name
     * @return ($name is 'checkbox_field' ? \Nette\Forms\Controls\Checkbox : ($name is 'select_field' ? \Nette\Forms\Controls\SelectBox : ($name is 'submit_button' ? \Nette\Forms\Controls\SubmitButton : IComponent)))
     */
    public function offsetGet($name): IComponent
    {
        return parent::offsetGet($name);
    }
}
```

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/77

### DIFF
--- a/src/Collector/FormCollector.php
+++ b/src/Collector/FormCollector.php
@@ -86,6 +86,7 @@ final class FormCollector implements Collector
             return null;
         }
 
+        // TODO find real form class and use it (e.g. $form = new Form())
         $formClassReflection = (new BetterReflection())->reflector()->reflectClass('Nette\Forms\Form');
 
         $formFields = [];
--- a/src/Collector/FormCollector.php
+++ b/src/Collector/FormCollector.php
@@ -148,7 +148,11 @@ final class FormCollector implements Collector
     {
         $methodCall = null;
         if ($expression->expr instanceof MethodCall) {
-            $methodCall = $expression->expr;
+            if ($expression->expr->var instanceof MethodCall) {
+                $methodCall = $expression->expr->var;
+            } else {
+                $methodCall = $expression->expr;
+            }
         } elseif ($expression->expr instanceof Assign && $expression->expr->expr instanceof MethodCall) {
             $methodCall = $expression->expr->expr;
         }
--- a/src/Collector/FormCollector.php
+++ b/src/Collector/FormCollector.php
@@ -1,0 +1,156 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Collector;
+
+use Efabrica\PHPStanLatte\Collector\ValueObject\CollectedForm;
+use Efabrica\PHPStanLatte\Collector\ValueObject\CollectedFormField;
+use Efabrica\PHPStanLatte\Resolver\NameResolver\NameResolver;
+use Efabrica\PHPStanLatte\Resolver\ValueResolver\ValueResolver;
+use PhpParser\Node;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Expression;
+use PHPStan\Analyser\Scope;
+use PHPStan\BetterReflection\BetterReflection;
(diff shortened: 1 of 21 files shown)


src/Collector/ValueObject/CollectedForm.php
```diff
@@ -0,0 +1,88 @@
...
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Collector\ValueObject;
+
+/**
+ * @phpstan-import-type CollectedFormFieldArray from CollectedFormField
+ * @phpstan-type CollectedFormArray array{className: class-string, methodName: string, name: string, formFields: CollectedFormFieldArray[]}
+ */
+final class CollectedForm
```
MartinMystikJonas, line 11: Do we really need specialised type? Is this Forms-specific? Or could there be other Components with fields. I am thinking if it would net be better to extend CollectedComponent to hold fields instead of new value object

src/Compiler/LatteToPhpCompiler.php
```diff
@@ -94,6 +93,69 @@ public function compileFile(?string $actualClass, string $templatePath, array $v
...
+                            new Name('parent'),
+                            new Identifier('offsetGet'),
+                            [
+                                new Arg(new NodeVariable('name')),
+                            ]
+                        )
+                    ),
+                ])
+                ->makePublic()
+                ->setReturnType($componentType);
```
MartinMystikJonas, line 145: This is really ugly :-D I would try to came up with better solution. But it is good enough.

src/Collector/FormCollector.php
```diff
@@ -0,0 +1,156 @@
+<?php
```
MartinMystikJonas, line 1: How much of this code is duplicated with componentFinder? Maybe create Abstract parent?

src/Compiler/LatteToPhpCompiler.php
MartinMystikJonas: maybe we can use array shapes for this? Or make it as properties and replace array access with property access? `...->form["field"]` to `...->form->field`

src/Collector/FormCollector.php
lulco: I think it's just finding of createComponent method

src/Compiler/LatteToPhpCompiler.php
lulco: Yes it is, this whole PR is just POC if we can handle forms :) there are lot of str / preg replaces etc.

src/Collector/ValueObject/CollectedForm.php
lulco: not all Components have fields, I think it is Form-specific

src/Compiler/LatteToPhpCompiler.php
lulco: Yes we probably can change all `end($formStack)["something"]` to $form->something and change conditional return type to `@property-read` comment in fake Form class

lulco:
Fuck it looks different in latte 3 :'(

src/Collector/ValueObject/CollectedForm.php
MartinMystikJonas: Well it could be optional property in Component. I would prefer to treat it as extended component - it would simplify code because we would not have to duplicate very similar collection/storing/finding with froms and components and passing them around etc. it owuld be just special version of Component

src/Compiler/LatteToPhpCompiler.php
MartinMystikJonas: I would preffer it that way in final version

lulco:
I would like to merge this as is, and create separate PR with Visitors and Class properties. Just, there is so rapid development from you side I had to do many changes :) 

This concept works, we just need nicer code

PR #99, opened 2022-12-03
Ignore some phpstan/strict-rules in compiled templates

MartinMystikJonas:


lulco:
Hmm something has changed?

MartinMystikJonas:
Weird - re run of tests on main passes

MartinMystikJonas:
I was unescaped `|`

lulco (approved):

PR #103, opened 2022-12-04
Collect vars passed as parameter #2 to render methods

MartinMystikJonas:
- variables passed to second argument of render() methods are collected (only arrays for now objects will be done in future)
- all calls to render template (included templates and calls to render() in PHP) are collected in same way (to avoid huge code diplication for very similar concept)
- when tempalte paths are evaluated then explicitly given path in first argument of render() method has priority, only if not provided or provided as null paths collected form setFile() are used (matches Latte behaviour)
- resolve calls to render() also in Presenters


src/LatteTemplateResolver/LatteTemplateResolverResult.php
```diff
@@ -44,7 +49,7 @@ public function getErrors(): array
 
     public function addTemplate(Template $template): void
     {
-        $this->templates[] = $template;
```
MartinMystikJonas, line 47: This prevents adding same template with exactly same context twice

src/Collector/Finder/MethodCallFinder.php
```diff
@@ -17,17 +17,17 @@ final class MethodCallFinder
     /**
      * @var array<string, array<string, array<string, string[]>>>
      */
-    private array $collectedMethodCalled;
```
MartinMystikJonas, line 20: To avoid uninitaliazed property errors when collected data are empty

src/Collector/Finder/TemplatePathFinder.php
```diff
@@ -35,16 +37,13 @@ public function __construct(CollectedDataNode $collectedDataNode, MethodCallFind
             if (!isset($this->collectedTemplatePaths[$className][$methodName])) {
                 $this->collectedTemplatePaths[$className][$methodName] = [];
             }
-            $templatePath = $collectedTemplatePath->getTemplatePath();
```
MartinMystikJonas, line 38: MOved to apth resolver because it is used in multiple places and is closely related to resolver wildcard mechanism

src/Collector/ValueObject/CollectedTemplateRender.php
```diff
@@ -0,0 +1,126 @@
...
+namespace Efabrica\PHPStanLatte\Collector\ValueObject;
+
+use Efabrica\PHPStanLatte\Template\Variable;
+use PHPStan\PhpDoc\TypeStringResolver;
+use PHPStan\ShouldNotHappenException;
+
+/**
+ * @phpstan-type CollectedTemplateRenderArray array{templatePath: string|bool|null, variables: array<array{variableName: string, variableType: string}>, className: string, methodName: string, file: string, line: int}
+ */
+final class CollectedTemplateRender
```
MartinMystikJonas, line 14: This class representr render call to template. Either from render() call in PHP or from including tempalte in latte.

src/Rule/LatteTemplatesRule.php
```diff
@@ -159,24 +154,24 @@ private function analyseTemplates(array $templates, Scope $scope, array &$errors
...
-            foreach ($collectedDataList as $collectedData) {
-                if ($collectedData->getCollectorType() !== IncludePathCollector::class) {
-                    continue;
-                }
-                /** @phpstan-var CollectedIncludePathArray[] $dataList */
-                $dataList = $collectedData->getData();
-                foreach ($dataList as $data) {
-                    $collectedIncludedPath = CollectedIncludePath::fromArray($data, $this->typeStringResolver);
+            $collectedDataList = $fileAnalyserResult->getCollectedData();
+            foreach ($this->templateRenderCollector->extractCollectedData($fileAnalyserResult->getCollectedData()) as $collectedTemplateRender) {
```
MartinMystikJonas, line 159: extraction of collected data moved to relevant collector to keep it together and hale Rule class cleaner

src/Collector/TemplateRenderCollector.php
```diff
@@ -0,0 +1,239 @@
...
+        if ($argument === null) {
+            return [];
+        }
+
+        $argumentType = $scope->getType($argument);
+
+        // todo support object (tempalte type)
+
+        $variables = [];
+        if ($argumentType instanceof ConstantArrayType) {
```
MartinMystikJonas, line 207: Use PHPStan mechnism of detecting array shapes to support wider variaty of expressions (like array assigned to variable)

src/Collector/TemplateRenderCollector.php
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        $functionName = $scope->getFunctionName();
+        if ($functionName === null) {
+            return null;
+        }
+
+        $currentClassType = new ObjectType($classReflection->getName());
+
+        if ($currentClassType->isInstanceOf('Latte\Runtime\Template')->yes()) {
```
MartinMystikJonas, line 80: Detects collector context - could be in future extended to more contexts that triggers render of template (like currently unsupported call to sendTempalte($file) in Presenters

MartinMystikJonas:
Looks like i used festure from newest PHPStan than current lowest? Rewrite or bump min PHPStan version?

lulco:
> Looks like i used festure from newest PHPStan than current lowest? Rewrite or bump min PHPStan version?

Bump

lulco (approved):
I think, we should add more tests.

MartinMystikJonas:
Any specific cases in mind?

lulco:
It seems these added tests have no errors. There should be some.

MartinMystikJonas:
I was checking just for existence of defined variables I will add some failing errors for unknown vars

lulco:
yes that's what I mean. Thanks

Issue #97, opened 2022-12-03
Custom filter registration


jakubvojacek:
Hello

do we have to manually register filters or is it enough to provide the engine bootstrap file? 

```
...
engineBootstrap: phpstan.latte.engine.php
...
```

```php
<?php

$container = require_once "app/bootstrap.php";


return $container->getService('latte.templateFactory')->createTemplate()->getLatte();
```

But phpstan still complaints about all my custom filters. I checked `->getFilters()` on the `getLatte()` and the filters were returned therefore I believe phpstan should be aware of them. But apparently not

Since this extension is still in very active development, is it too early to report this? Should I wait some more weeks?

Thanks

MartinMystikJonas:
Thanks for report.  What version of Latte do you use? For latte 3 it should work for filters added by extensions. For latte 2 there is problem that we cannot get filter signatures we need just filter names. I have some idea how to get to this info using reflection but it is not yet implemented.

jakubvojacek:
3.0.4, the very latest. The filters are added in config.neon

```
services: 
	installLatteFilters: Motv\FrontModule\Lib\InstallLatteFilterMacros
	latte.latteFactory:
		setup:
			- @installLatteFilters::install(@self)
```

and 

```php
...
public function install(Latte\Engine $template): void
	{
		$template->addFilter('markdown', static function (string $text): string {
			$parser = new \Parsedown();

			return $parser->parse($text);
		});
....
```

MartinMystikJonas:
Ok this is not supported yet in Latte3. Only filters added by extensions.

MartinMystikJonas:
For now you can either add filters in config or rewrite you code to use Latte 3 extension for filter registration. I will try to solve it for directly added filters in the future.

jakubvojacek:
> For now you can either add filters in config or rewrite you code to use Latte 3 extension for filter registration

not sure which part of the docs describes this, is there some github sample in some project or nette docs? I thought I was doing it correctly - its how I found it described by dg in some thread on nette forums

Thanks

MartinMystikJonas:
Documentation: https://latte.nette.org/cs/creating-extension
Tutorial by @dg: https://forum.nette.org/cs/35241-latte-3-a-extension-pro-kazdy-webovy-projekt

jakubvojacek:
Thanks, the forum post seems like a spot-on. I will rewrite my code according to that.

[...]

PR #105, opened 2022-12-05
Support closure filters

### LINKED ISSUE #57: Closure filters
Support for checking filters added as closures. Would require adding fictional method to template with same signature as closure and replacing filter calls with calls to this method.

### LINKED ISSUE #60: Latte 2 - detect non default filters set in Engine and add them too
Is it possible to ask Engine for filter type in Latte 2? If not at least detect that filter exists.

### LINKED ISSUE #61: Detect use of undefined filters
Add LateCimpileError when filter is unknown here https://github.com/efabrica-team/phpstan-latte/blob/fe711dac0ad0d88f19a14db8be15d19a8d2ecaa8/src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php#L89

### LINKED ISSUE #100: `Cannot use object of type Closure as array` when using filters
Hello

I get 

```
 ------ -------------------------------------------------------------------------------------- 
  Line   PersonsGrid.latte rendered from Motv\FrontModule\Components\Vods\PersonsGrid::render  
 ------ -------------------------------------------------------------------------------------- 
  -1     Cannot use object of type Closure as array                                            
 ------ -------------------------------------------------------------------------------------- 
```

on a very simple file

```
{if $person !== null}
	{$person->getBirthday()|datetime}
{/if}
```

I cannot share the generated file because none is generated actually. 

The filter `|datetime` is register as latte3 extension and defined as

```php
/** @param mixed $datetime */
	public static function formatDateTime($datetime): string
	{
		if ($datetime === null) {
			return 'N/A';
		}

		if ($datetime instanceof \DateTime || $datetime instanceof \DateTimeImmutable) {
			$datetime = $datetime->format('c');
		}

		if (\is_string($datetime)) {
			$datetime = strtotime($datetime);
		}

		return date('D, j M Y G:i:s', $datetime);
	}


...
                         'datetime' => static function ($date): string {
				return Helper::formatDateTime($date);
			},
...
```

MartinMystikJonas:
Resolves #57 
Resolves #60 
Resolves #61
Resolver #97 
Resolves #100

### DIFF
--- a/tests/Rule/LatteTemplatesRule/EngineBootstrap/LatteTemplatesRuleForEngineBootstrapTest.php
+++ b/tests/Rule/LatteTemplatesRule/EngineBootstrap/LatteTemplatesRuleForEngineBootstrapTest.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\PresenterWithoutModule;
+
+use Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\LatteTemplatesRuleTest;
+
+final class LatteTemplatesRuleForEngineBootstrapTest extends LatteTemplatesRuleTest
+{
+    public static function getAdditionalConfigFiles(): array
+    {
+        return [
+            __DIR__ . '/../../../../rules.neon',
+            __DIR__ . '/../../../config.neon',
+            __DIR__ . '/config.neon',
+        ];
+    }
+
+    public function testFilters(): void
+    {
+        $this->analyse([__DIR__ . '/Fixtures/FiltersPresenter.php'], [
+            [
+                'Trying to invoke mixed but it\'s not a callable.',
+                2,
+                'default.latte',
+            ],
+            [
+                'Undefined latte filter "nonExistingFilter".',
+                2,
+                'default.latte',
+                'Register it in phpstan.neon: parameters > latte > filters. See https://github.com/efabrica-team/phpstan-latte#filters',
+            ],
+            [
+                'Closure invoked with 1 parameter, 0 required.',
+                3,
+                'default.latte',
+            ],
+            [
+                'Parameter #1 $ of closure expects int, string given.',
+                4,
+                'default.latte',
+            ],
+        ]);
+    }
+}
--- a/tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
+++ b/tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
@@ -1,0 +1,12 @@
+<?php
+
+use Latte\Engine;
+
+$engine = new Engine();
+$engine->addFilter('closure', function () {
+});
+$engine->addFilter('closureWithIntParam', function (int $param) {
+});
+$engine->addFilter('closureWithStringParam', function (string $param) {
+});
+return $engine;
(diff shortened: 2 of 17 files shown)


tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+use Latte\Engine;
+
+$engine = new Engine();
+$engine->addFilter('closure', function () {
```
lulco, line 6: will this check closure parameters? for example here there is no parameter in function() but $title is given but no error is reported

tests/Rule/LatteTemplatesRule/EngineBootstrap/LatteTemplatesRuleForEngineBootstrapTest.php
```diff
@@ -0,0 +1,36 @@
...
+            __DIR__ . '/../../../config.neon',
+            __DIR__ . '/config.neon',
+        ];
+    }
+
+    public function testFilters(): void
+    {
+        $this->analyse([__DIR__ . '/Fixtures/FiltersPresenter.php'], [
+            [
+                'Trying to invoke mixed but it\'s not a callable.',
```
lulco, line 24: do we need this error? the second one is nice and I like it :)

lulco (approved):
Some tests are failing, please fix it

tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
MartinMystikJonas: It depends on how PHPStan converts closure to ClosureType and how it evaluates if closure call matches that type. I think it specifies params when provided but I am not sure how it handleds when params are ommited

tests/Rule/LatteTemplatesRule/EngineBootstrap/LatteTemplatesRuleForEngineBootstrapTest.php
MartinMystikJonas: First error is side effect. We cannot ignore it because exactly same error could be triggered by user code. But maybe we can generate different code that will not trigger first error when filter is unknown.

tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
MartinMystikJonas: Hmm just checked it and for some reson it is always build just like \Closure without types. I will investigate why

tests/Rule/LatteTemplatesRule/EngineBootstrap/LatteTemplatesRuleForEngineBootstrapTest.php
lulco: Yeah, we wrote about it before. I recall... So if we can't ignore it, then we have to leave with it :(

lulco (approved):

tests/Rule/LatteTemplatesRule/EngineBootstrap/LatteTemplatesRuleForEngineBootstrapTest.php
MartinMystikJonas: I will try to solve it in the future

PR #108, opened 2022-12-05
Collect vars passed as parameter #2 to render methods - object support

### LINKED ISSUE #107: Support object passed to argument #2 or render

MartinMystikJonas:
Resolves #107

### DIFF
--- a/src/Collector/TemplateRenderCollector.php
+++ b/src/Collector/TemplateRenderCollector.php
@@ -200,8 +200,9 @@ final class TemplateRenderCollector implements Collector
         }
 
         $argumentType = $scope->getType($argument);
-
-        // todo support object (tempalte type)
+        if ($argumentType instanceof ObjectType) {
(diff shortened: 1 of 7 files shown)


lulco:
I don't know how this exactly works, but in this test cases I would expect an app to throw error about unitialized properties which are accessed in latte. PHPStan should also throw this error

MartinMystikJonas:
We do not have this information available in static anylysis. We are limited by PHPStan capabilities here.

lulco:
So we basically don't test anything for renders with object as 2nd parameter. Template class can say there are fields $a, $b, $c and if developer will not set them, phpstan latte will throw no error.

MartinMystikJonas:
Well recommended use case are constructor promoted properties when it will be handled on PHP level. But when developer passes object with uninitialised properties it cannot be detected by PHPStan.

lulco (approved):
OK we can merge it now. We will see if there will be some issues

PR #115, opened 2022-12-06
Added type for Form collector and use it in compiled template

### LINKED ISSUE #101: Custom formRender and false positive errors such as `renderPair`
Hello

we use custom formRender (extending `Nette\Forms\Rendering` implementing `Nette\Forms\FormRenderer`)

It has methods such as `renderPair, renderControls, ...`

We use it directly in lattes sometimes, eg 

```
{$form->getRenderer()->renderControls($form->getGroup('General'), true) |noescape}
```

or 

```
{$form->getRenderer()->renderPair($formContainer->getComponent('channels_audio_pid')) |noescape}
```

which results in phpstan complaining about

```
  133    Call to an undefined method Nette\Forms\FormRenderer::renderPair().                                                   
  136    Call to an undefined method Nette\Forms\FormRenderer::renderPair().                                                   
  139    Call to an undefined method Nette\Forms\FormRenderer::renderPair().                                                   
  144    Call to an undefined method Nette\Forms\FormRenderer::renderPair().                                                   
  147    Call to an undefined method Nette\Forms\FormRenderer::renderPair().                                                   
  150    Call to an undefined method Nette\Forms\FormRenderer::renderPair().  
....
```

I can add these errors into my phpstan ignore list. But I was wondering if this should be done at the phpstan-latte level? Since `Nette\Forms\RenderingDefaultFormRenderer` which is used by default implements all these methods?

Thanks

lulco:
Resolves #101

### DIFF
--- a/src/Collector/FormCollector.php
+++ b/src/Collector/FormCollector.php
@@ -143,6 +143,7 @@ final class FormCollector extends AbstractCollector
             $classReflection->getName(),
             '',
             $formName,
+            $returnType,
             $formFields
         ));
     }
--- a/src/Collector/ValueObject/CollectedForm.php
+++ b/src/Collector/ValueObject/CollectedForm.php
@@ -4,9 +4,12 @@ declare(strict_types=1);
 
 namespace Efabrica\PHPStanLatte\Collector\ValueObject;
 
+use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\Type;
+
 /**
  * @phpstan-import-type CollectedFormFieldArray from CollectedFormField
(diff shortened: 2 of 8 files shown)


MartinMystikJonas (approved):

MartinMystikJonas:
Could resulting type be UnionType? It is probaly edge case but should he coveret to be sure to not generate invalid php.

MartinMystikJonas:
Or maybe all cases when it is not ObjectType (could be mixed with missing type hint)

lulco:
At the moment it will not pass through this condition:
```
$returnType->isInstanceOf('Nette\Forms\Form')->yes()
```

But yes, in theory it can be any type

PR #117, opened 2022-12-06
Custom form methods

lulco:
@MartinMystikJonas any idea how to get information about this custom method? (it is added via Container::extensionMethod)

This:
```
@method TextInput addCustomText(string $name, ?string $label = null, ?int $cols = null, ?int $maxLength = null)
```
was added for sake of phpstan, and it can read it. But I have no idea when, where and how :) I want to use it in FormCollector because now all such fields are skipped.


MartinMystikJonas:
I think methods added by @method should be visible to PHPStan reflection. Instead of using BetterReflection directly try to inject ReflectionProvider (or some similar name I do not know exact name for sure) service of PHPStan.

MartinMystikJonas:
https://phpstan.org/developing-extensions/reflection Check this part of docs. I think it is what you need.

MartinMystikJonas:
When I think about it we probably should use it in most (all) places instead of BetterReflection to take advantage of additional info processed by PHPStan.

lulco:
Thanks I’ll check it. And if it will be better, I agree we can replace it

lulco:
It's little bit ugly (try catch getMethod(), getVariants() and then getReturnType()), but works like a charm! 🎊

MartinMystikJonas:
Isnt there hasMethod in class Reflection? It would be better than try catch

lulco:
hasMethod used

MartinMystikJonas (approved):

PR #118, opened 2022-12-06
Improved detection of terminating and output calls

### LINKED ISSUE #96: Improve detection of early terminated presenter actions
See https://github.com/phpstan/phpstan/discussions/8452#discussioncomment-4297673

MartinMystikJonas:
- detects early terminating calls (in method itself or in any called method)
- detect always terminated methods (all paths throws exception or otherwise ends) (in method itself or in any called method)
- detect calls to functions with return never (in method itself or in any called method)
- detect outputing calls in Control render methods (echo, print. print_r, var_dump) and do not report unresolved latte template when render methods outputs directly
- 
Future improvement: detect if evaluated calls happens always or conditionally - now it will not report resolve error even when some paths can lead to attempt to tender template - would require information if function call is guaranteed or optional in CollectedMethodCall based on context of call

Resolves #96


src/Collector/ComponentCollector.php
```diff
@@ -100,7 +100,7 @@ private function findAddComponent(MethodCall $node, Scope $scope, ClassReflectio
     {
         // TODO check if caller class is control / presenter
 
-        if ($this->nameResolver->resolve($node->name) !== 'addComponent') {
+        if ($this->nameResolver->resolve($node) !== 'addComponent') {
```
lulco, line 103: Will this work? How? We can’t know if user wants method name or expression (variable or another method call name)

lulco (approved):

src/Collector/ComponentCollector.php
MartinMystikJonas: When you want name of something expcific you can pass more specific node. But by default it returns name of the node - it was in all calls in current code.

PR #121, opened 2022-12-06
Cleanup visitor

lulco:
Fix for error `Expression "$form" on a separate line does not do anything.`

after changes, there can stay expression where only variable is placed, e.g.
```php
$form;
```
we can remove it.

This CleanupVisitor can be used also for other use cases to cleanup.

### DIFF
--- a/src/Compiler/Compiler/AbstractCompiler.php
+++ b/src/Compiler/Compiler/AbstractCompiler.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\Compiler\Compiler;
 
 use Latte\Engine;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\IntegerType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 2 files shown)


src/Compiler/Compiler/AbstractCompiler.php
```diff
@@ -71,7 +73,7 @@ protected function generateTypes(string $className, array $types): string
             } else {
                 $type = is_object($value) ? get_class($value) : gettype($value);
             }
-            $phpCode .= "* @property-read {$type} \${$name}\n";
```
lulco, line 74: changed `@property-read` to `@property` because of formsStack where write is needed

src/Compiler/LatteToPhpCompiler.php
```diff
@@ -213,6 +217,31 @@ private function addExtractParams(string $phpContent): string
         return $this->printerStandard->prettyPrintFile($phpStmts);
     }
 
+    private function cleanup(string $phpContent): string
+    {
+        $phpStmts = $this->findNodes($phpContent);
+
+        $nodeTraverser = new NodeTraverser();
+        $cleanupVisitor = new class extends NodeVisitorAbstract
```
MartinMystikJonas, line 225: I would create normal class

lulco: Sure, I will do some refactoring - move all these visitors to some “postprocessor”

Issue #85, opened 2022-11-30
Collect data from nonanalyzed but used files


lulco:
e.g. application calls some classes from vendor and there are some variables assigned, components created etc. We have to analyze these files and merge collected data together

lulco:
Resolved by https://github.com/efabrica-team/phpstan-latte/pull/124

PR #124, opened 2022-12-07
Collect additional data from related files

lulco:
Related to: https://github.com/efabrica-team/phpstan-latte/issues/85

This is first part of solution. In the middle of it I realized that we need also files of classes created, called etc - will be done as follow up for this PR. Probably we just extend ParentCollector to collect some data from other nodes.

As you can see, thanks to this approvement, we can remove some classes (ParentPresenter) from tests. In testComponents we still have to keep called classes of Controls


lulco:
@MartinMystikJonas I have no idea what you did with collectors :D need help here to solve phpstan errors ;)

MartinMystikJonas:
You just lost generics template on new interface. Add this to pass type info:
```php
/**
 * @template N of Node
 * @template T of CollectedValueObject
 * @template A of array
 * @implements Collector<N, ?A[]>
 */
interface PHPStanLatteCollectorInterface extends Collector
```
And in AbstractCollector change implements to corrent class:
```
@implements PHPStanLatteCollectorInterface<N, ?A[]>
```

PHPStan does not allow to pass generics without specifying it in every class/interface in hierarchy. SO we cannot just skip it in `PHPStanLatteCollectorInterface`

MartinMystikJonas:
At first sight looks good but I will do proper PR later

src/Collector/ParentCollector.php
```diff
@@ -0,0 +1,35 @@
...
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\ExecutionEndNode;
+
+final class ParentCollector implements PHPStanLatteCollectorInterface
+{
+    public function getNodeType(): string
+    {
+        return ExecutionEndNode::class;
```
MartinMystikJonas: Wrong node. This will run on end of each function and loop. Use InClassNode

src/Rule/LatteTemplatesRule.php
```diff
@@ -101,6 +113,51 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
...
+        }
+
+        $newCollectedDataNode = new CollectedDataNode(array_merge(...$collectedData));
+        return $this->collectAdditionalData($newCollectedDataNode);
+    }
+
+    /**
+     * @return CollectedData[]
+     */
+    private function createCollectedDataFromNode(CollectedDataNode $collectedDataNode): array
```
MartinMystikJonas: I would move this somewhere to Collectors anmespace. Maybe specialised service?

src/Rule/LatteTemplatesRule.php
```diff
@@ -101,6 +113,51 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
         return $errors;
     }
 
+    private function collectAdditionalData(CollectedDataNode $collectedDataNode): CollectedDataNode
+    {
+        $collectedParents = $collectedDataNode->get(ParentCollector::class);
```
MartinMystikJonas: this logic of extracting collected data should be in AbstractCollector/ParentCollector similar to how extractCollectedData is there - keep extracting together wit collection to simplify changes in collected data format

src/Collector/ParentCollector.php
```diff
@@ -0,0 +1,35 @@
...
+        if ($classReflection === null) {
+            return null;
+        }
+
+        $relatedFiles = [];
+        foreach ($classReflection->getParents() as $parentClassReflection) {
+            $relatedFiles[] = $parentClassReflection->getFileName();
+        }
+
+        return array_unique(array_filter($relatedFiles));
```
MartinMystikJonas: Maybe create ValueObject something like CollectedAdditionalFile and use it instead of string. To keep all collectors behave in same way. I would not create CollectedParent because we would have more collectors collecting this.

src/Collector/PHPStanLatteCollectorInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Collector;
+
+use PHPStan\Collectors\Collector;
+
+interface PHPStanLatteCollectorInterface extends Collector
```
MartinMystikJonas: add loadData, extracCollectedData and processNode signatures here

src/Collector/ParentCollector.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public function processNode(Node $node, Scope $scope)
+    {
+        $classReflection = $scope->getClassReflection();
+        if ($classReflection === null) {
+            return null;
+        }
+
+        $relatedFiles = [];
+        foreach ($classReflection->getParents() as $parentClassReflection) {
```
MartinMystikJonas: Do we need parents of all classes or only classes matched by any of resolvers (Controls, Presenters)?

src/Rule/LatteTemplatesRule.php
```diff
@@ -101,6 +113,51 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
...
+        $processedFiles = array_unique(array_merge(array_keys($collectedParents)));
+        $relatedFiles = array_unique(array_merge(...array_merge(...array_values($collectedParents))));
+        $newFilesToCheck = array_diff($relatedFiles, $processedFiles);
```
MartinMystikJonas: I am bit lost how this check works under the hood. How do you detect which filesare new and which was already collected?

lulco:
I think I did it, but didn’t work. I’ll try it again, Maybe I forgot something

MartinMystikJonas (changes requested):

MartinMystikJonas:
I thiunk we should also check all call-chains starting at resolved nodes. Get resolved nodes - get all directly and indirectly called methods and collect call targets that points to unanalysed classes

MartinMystikJonas:
This might also be quite performance/memory consumption boost if we first collect just resolved nodes and then use other collectors only on needed files in analysed paths. Question is if multiple small runs on subset would be better that single big one on evwryrhing.

src/Collector/ParentCollector.php
lulco: Thanks, I used Class_ and ClassLike but they dont have classReflection, I'll try InClassNode

src/Collector/ParentCollector.php
lulco: That's great idea - as we wrote on slack, we will collect just classes from resolvers and then we will collect other data

lulco:
@MartinMystikJonas I rewrite it to collect ResolvedNodes first and then other data but after adding collect called classes, it slows down.

Also I have to skip all collector result tests, because now we are collecting only resolved nodes (only this collector is in getCollectors())

MartinMystikJonas:
How significant is slow down?

Collector result tests would be eas to fix after you move collectAdditionalData to separate service - test can call this service too to get required data.

lulco:
when I wrote about slow down, on local machine it was 2 minutes, here test lasts 6+ minutes and that's tooooo much.

I'm thinking about this again. We don’t need to collect data from all codebase (I think we don’t care about method calls inside vendor/nette etc.) so we can add parameter to config where the additional data should be collected. I'll try to implement it.

src/Collector/ResolvedNodeCollector.php
```diff
@@ -15,7 +15,7 @@
  * @phpstan-import-type CollectedResolvedNodeArray from CollectedResolvedNode
  * @extends AbstractCollector<Node, CollectedResolvedNode, CollectedResolvedNodeArray>
  */
-final class ResolvedNodeCollector extends AbstractCollector
+final class ResolvedNodeCollector extends AbstractCollector implements PHPStanLatteCollectorInterface
```
MartinMystikJonas: Why do not add this implements directly to AbstractCollector?

src/Collector/RelatedFilesCollector.php
```diff
@@ -0,0 +1,72 @@
...
+                if (!$classReflection->isInterface() && !$classReflection->isTrait()) {
+                    $relatedFiles[] = $classReflection->getFileName();
+                }
+            }
+        } elseif ($node instanceof CallLike) {
+            $calledClassName = $this->calledClassResolver->resolve($node, $scope);
+            if ($calledClassName !== null) {
+                $classReflection = $this->reflectionProvider->getClass($calledClassName);
+                if (!$classReflection->isInterface() && !$classReflection->isTrait()) {
+                    $relatedFiles[] = $classReflection->getFileName();
```
MartinMystikJonas: I think there should be also check that stops searching for additional files on external library classes like what is in MethodCallCollector (https://github.com/efabrica-team/phpstan-latte/blob/1fde35416fdb4e40a88b8ff40c221d180dfceebd/src/Collector/MethodCallCollector.php#L126) to stop unnecessary collection inside Netta/Latte. This might be reason for performance degradation. 

This check for external class should probably be created as resolver in Resolver\TypeREsolver and used in both MethodCallCollector and RelatedFilesCollector. 

Also this check might need to be more strict - why do not exclude whole Latte and Nette? 

Also I thnik we might want to check call arguments - if call is not in same class, do not pass $this and do not pass Template then this call cannot change latte related state so we can ignore it.

MartinMystikJonas:
Configurable ignore is good idea. Put nette/latte and maybe some other libs there and allow user to configure. 

Also I wrote this idea in rewiew comment: Also I thnik we might want to check call arguments - if call is not in same class or ancestor, do not pass $this and do not pass Template then this call cannot change latte related state so we can ignore it.

It could greatly reduce checked calls also in app.

lulco:
> Configurable ignore is good idea. Put nette/latte and maybe some other libs there and allow user to configure.
> 
> Also I wrote this idea in rewiew comment: Also I thnik we might want to check call arguments - if call is not in same class or ancestor, do not pass $this and do not pass Template then this call cannot change latte related state so we can ignore it.
> 
> It could greatly reduce checked calls also in app.

I was thinking about white list. In my case it would be one or two directories. Actual setup stops propagation on selected classes from nette, but for example usage of guzzle http client will analyse also its calls. And there are several such packages which are called directly from app.

src/Collector/ResolvedNodeCollector.php
lulco: In this version I just rewrite collectors to collect only resolved nodes and rest collectors are called only on them.

MartinMystikJonas:
It could be unexpected for user if they do not know about this option. I think we should do check of everything that can change latte state by default and allow user to configure for better performance. Imho whitelist is good idea as long as default is everything except nette/latte is whitelisted.

MartinMystikJonas:
Analyse of guzzle calls and similar and also latte unrelated internal calls in app could be avoided by that argument checking. Unless called method can get access to template somehow (got it as param, have access to it through property) it cannot influence what is passed to template and we can ignore it.

MartinMystikJonas:
For every method call check if called class is current class or any ancestor. Id it is check it. If not then iterate over arguments and for each of then ask scope to get type if type is current class or Template check it. If argument call type is mixed check also type deckared in called method signature. If all is unknown then check it just to be sure.

MartinMystikJonas:
This would also fix some false results. For example when Control calls method of sub component to chenge it's state then all template assigns made by sub component to sub component tempalte are currently interpreted as assigned to calling component template too.

src/Collector/ResolvedNodeCollector.php
MartinMystikJonas: Yes but I do no see reason behind this? If all AbstractCollector offsprings implements this interface why do not add it to AbstractCillector? What am I missing?

src/Collector/RelatedFilesCollector.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public function processNode(Node $node, Scope $scope): ?array
+    {
+        $relatedFiles = [];
+        if ($node instanceof InClassNode) {
+            $classReflection = $scope->getClassReflection();
+            if ($classReflection !== null) {
+                foreach ($classReflection->getParents() as $parentClassReflection) {
+                    $relatedFiles[] = $parentClassReflection->getFileName();
```
MartinMystikJonas: Use relatedbfile also as key to prevent adding it multiple times

MartinMystikJonas:
We probably also need to implement some kind of cache for collector results. I think now we avoided phpstan internal results cache for collector results.

lulco:
@MartinMystikJonas I'm done, please approve

MartinMystikJonas (approved):

MartinMystikJonas:
Approved.

MartinMystikJonas:
Later add docs about new configuration option (but I would do it after we merge changed structure of docs)

PR #125, opened 2022-12-07
Tests refactoring

MartinMystikJonas:
Series of improvements to make writing and maintainiung tests much easier

- fixtures are autodetected - no need to add them manually
- fixtures are self-containing - everything fixture needs including expected errors is packed in fixture
- new fixtures can be created simply by copying exissting fixture and changing namespace
- expected errors and collector results are parsed from source files where it is present in form of special comments (no need to manually write/update line numbers when source changes)
```latte
{control nonExistingControl} {* ERROR: Component with name "nonExistingControl" probably doesn't exist. *}
```
```php
    public function renderDefaultVars(): void
    {
        $this->template->setFile(__DIR__ . '/defaultVars.latte');
        // COLLECT: TEMPLATE defaultVars.latte SomeControl::renderDefaultVars ["presenter","control","a","b"] []
        $this->template->render(null, ['a' => 'a', 'b' => 'b']);
    }
```
- new tested parts can we easily created by copy-pasting existing parts of code including expected errors/collects and modifying them
- support for adding multiple errors/collects for one line by prepending lines containing only expected errers/collects before line they should be expected on
```latte
        {* ERROR: Cannot access property $title on mixed. *}
        {* ERROR: Cannot access property $id on mixed. *}
        {$item->id} - {$item->title}
```
- errors that cannot be presented in code comments are written in special file `_expectedErrors.php`
```php
<?php

return [
    [
        'Template file "' . __DIR__ . '/invalid_file.latte" doesn\'t exist.',
        -1,
        'invalid_file.latte',
    ],
];
```


lulco (commented):
Sorry, I really don’t like it. I understand you spend a lot of time here, but this is not how I imagine tests in my package.

- you’re created your own syntax which have to be used
- you lost some informations (line number, included template parent)
- you are not testing “real world” classes and templates because they contain additional data (comments you used to collect errors etc.)

adding new test are no problem in actual setup, you can copy method too if you are not able to create it yourself

lines changes only id you add some variables / components etc before or between existing code - add it to the end of file - it is very important to have line in developer's hand because in compiled template it is just comment and in some versions of nette / latte it works differently

MartinMystikJonas:
Sorry to hear that because I think it would greatly simplify writing tests. For annotations tests I would need to create lots of modified fixtures and it is quite painful in current setup.

For your objections:
1) **you’re created your own syntax which have to be used** - Well yes but this convention can save lots of time specifying/updating things manually (it is inspired by https://github.com/shipmonk-rnd/phpstan-rules I developed PR for it and it was so nice to use I loved it)
2) **you lost some informations (line number, included template parent)** - No information is lost here I believe. It is just parsed from context where comment is written - exact line number is irrelevant - context matters. Otherwise it is exactly same as it was written manually
3) **you are not testing “real world” classes and templates because they contain additional data (comments you used to collect errors etc.)** - they are exactly the same as they were just with added annoations, no changes in class code structure to suit tests
4) **lines changes only id you add some variables / components etc before or between existing code** - in my case I would need to create lots of different fixtures with modifications all over the code (added/moved annotation), not just added content so line numbers would keep changung - that was my main motivation for this PR

Line numbers are parsed from original source file not from compiled file - there could be no lost of line number information by compilation.

But it is your package so decision is yours of course :-) I would be disappointed but I can live with it.

At least I would like some version where Fixtures are self-contained. Maybe just keep mechanism used by _expectedErrors.php?

Or can I keep it at least for annotations tests and keep them separated?

lulco:
> Sorry to hear that because I think it would greatly simplify writing tests. For annotations tests I would need to create lots of modified fixtures and it is quite painful in current setup.

Unfortunatelly, the same feeling I had yesterday with collectors after refactoring. Previous version was clear, each collector did its own tasks, there was no abstraction, no static methods. I just change one collector, and other can stay untouched. Now it is hell (for me). And I am afraid it will be the same with tests :)

> 
> For your objections:
> 
> 1. **you’re created your own syntax which have to be used** - Well yes but this convention can save lots of time specifying/updating things manually (it is inspired by https://github.com/shipmonk-rnd/phpstan-rules I developed PR for it and it was so nice to use I loved it)

I love automation where it seem reasonable. I just don't see everything I think I need here. For example:
- Where are additional php files configured? (in some tests - e.g. testComponents we need several files included) or do you test always everything?

> 2. **you lost some informations (line number, included template parent)** - No information is lost here I believe. It is just parsed from context where comment is written - exact line number is irrelevant - context matters. Otherwise it is exactly same as it was written manually

Exact line number is relevant imho - if you get error about error on line 10, but real error is on 9 or 11, it is very confusing.

> 3. **you are not testing “real world” classes and templates because they contain additional data (comments you used to collect errors etc.)** - they are exactly the same as they were just with added annoations, no changes in class code structure to suit tests

Those added annotations are the change which can be interpret differently (I know they are not now, but it is not the same code as in real apps and can change behavior of test unpredictably)

> 4. **lines changes only id you add some variables / components etc before or between existing code** - in my case I would need to create lots of different fixtures with modifications all over the code (added/moved annotation), not just added content so line numbers would keep changung - that was my main motivation for this PR
> 
> Line numbers are parsed from original source file not from compiled file - there could be no lost of line number information by compilation.

This is the problem, line numbers are parsed from compiled file, not from original source. There are several fixes and hacks in compilation to get line numbers on correct places to teach phpstan-latte where the code was in original source file. Without it we never found bus for example in nette/application 3.0 vs 3.1 where the behavior was changed (in older versions the line comment was in different place or wasn't there at all)

> 
> But it is your package so decision is yours of course :-) I would be disappointed but I can live with it.

I understand your feelings. It's just 100+ changes in diff and it is code I'll have to maintain in the future (even when you decide to leave the project)

> 
> At least I would like some version where Fixtures are self-contained. Maybe just keep mechanism used by _expectedErrors.php?

Well, errors are not together, I think it complicate things little bit.

> 
> Or can I keep it at least for annotations tests and keep them separated?

I think, yes. Create separate test fixtures for annotations and use this mechanism there. Maybe I will like it when I will use it later. I don't know now.

Thank you.

MartinMystikJonas:
> > Sorry to hear that because I think it would greatly simplify writing tests. For annotations tests I would need to create lots of modified fixtures and it is quite painful in current setup.
> 
> Unfortunatelly, the same feeling I had yesterday with collectors after refactoring. Previous version was clear, each collector did its own tasks, there was no abstraction, no static methods. I just change one collector, and other can stay untouched. Now it is hell (for me). And I am afraid it will be the same with tests :)
> 

Well my goal was exect opposite. This abstraction should simplify things. Maybe we can discuss how I ment it to work on slack. Might be just something I did not made clear enough. Only drawback is that collectors are meant to always use value objects - it is not really ready to collect plain strings - that might be the root cause of your pain.

> > For your objections:
> > 
> > 1. **you’re created your own syntax which have to be used** - Well yes but this convention can save lots of time specifying/updating things manually (it is inspired by https://github.com/shipmonk-rnd/phpstan-rules I developed PR for it and it was so nice to use I loved it)
> 
> I love automation where it seem reasonable. I just don't see everything I think I need here. For example:
> 
> * Where are additional php files configured? (in some tests - e.g. testComponents we need several files included) or do you test always everything?

Fixture analyses all files put directly into fixture folder (not recursively). If you want additional files that are not analysed you can either put them outside fixtures folder (if it is for example something that should be shared by multiple fixtures) or in subfolder in fixture (if it is relevant only to given fixture).

> > 2. **you lost some informations (line number, included template parent)** - No information is lost here I believe. It is just parsed from context where comment is written - exact line number is irrelevant - context matters. Otherwise it is exactly same as it was written manually
> 
> Exact line number is relevant imho - if you get error about error on line 10, but real error is on 9 or 11, it is very confusing.

IMHO it is relevant if error is raised on line where causing code is located. Exact numerical value of line is irrelevant if error is reported on causing line - and it will be if error annotation is written right next to causing code. even when surrounding code changes if causing code and error are kept together it will test that line number is reported correctly.

> > 3. **you are not testing “real world” classes and templates because they contain additional data (comments you used to collect errors etc.)** - they are exactly the same as they were just with added annoations, no changes in class code structure to suit tests
> 
> Those added annotations are the change which can be interpret differently (I know they are not now, but it is not the same code as in real apps and can change behavior of test unpredictably)

Not sure how it could change behaviour. From perspective of tested code it is just comment.

> > 4. **lines changes only id you add some variables / components etc before or between existing code** - in my case I would need to create lots of different fixtures with modifications all over the code (added/moved annotation), not just added content so line numbers would keep changung - that was my main motivation for this PR
> > 
> > Line numbers are parsed from original source file not from compiled file - there could be no lost of line number information by compilation.
> 
> This is the problem, line numbers are parsed from compiled file, not from original source. There are several fixes and hacks in compilation to get line numbers on correct places to teach phpstan-latte where the code was in original source file. Without it we never found bus for example in nette/application 3.0 vs 3.1 where the behavior was changed (in older versions the line comment was in different place or wasn't there at all)

Expected line numbers of errors are parsed directly from original source file (expected errors parser uses original source file not compiled version). And actual errors are processed by compiler/error builder. So if compiler/error builder do something wrong it will be detected because actual line and expected line would differ.

> > But it is your package so decision is yours of course :-) I would be disappointed but I can live with it.
> 
> I understand your feelings. It's just 100+ changes in diff and it is code I'll have to maintain in the future (even when you decide to leave the project)

As I said I would be disappointed but I completely understood. My hopes was that this would make code easier to miantain but your opinion is what matters, because you would be one who would have to miantain it in long term. I can just try to present good arguments :-)

> > At least I would like some version where Fixtures are self-contained. Maybe just keep mechanism used by _expectedErrors.php?
> 
> Well, errors are not together, I think it complicate things little bit.

Not sure what exactly do you mean by this. Errors from different fixtures should not be together imho ebcause fixtures should be self-contained. Now all error are in single file and with adding more and more fixtures it would be really long file eventually. Also adding new fixture now requires changes in multiple places - that is one thing I wanted to avoid - keep related code changes for usual use-cases as close as possible.
 
> > Or can I keep it at least for annotations tests and keep them separated?
> 
> I think, yes. Create separate test fixtures for annotations and use this mechanism there. Maybe I will like it when I will use it later. I don't know now.

As long as I can use it for annotations tests I am happy. It could save me several hours of manualy adjusting line numbers in these tests. It is not such big problem in tests we already have but when I started writing annotations tests it was real pain. One added annoation to class and I had to change all line numbers in that test expected errors.

> Thank you.

PR #126, opened 2022-12-08
Annotation @phpstan-latte-ignore

MartinMystikJonas:
Just first draft of annotations support with @phpstan-latte-ignore

Would appreciate feedback before I continue with other annotations and detailed tests


src/PhpDoc/LattePhpDoc.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Efabrica\PHPStanLatte\PhpDoc;
+
+final class LattePhpDoc {
```
lulco: please fix codestyle, { to new line etc

lulco (commented):
Looks good, just fix sniffer and other tests.

I have one question: after ignoring some methods, don't we have some extra untested latte files? Let's say I have actionDefault (ignored) and default.latte, will default.latte be marked as untested?

src/PhpDoc/LattePhpDoc.php
MartinMystikJonas: I will I did not care about code style yet it was just proof of concept at this point.

MartinMystikJonas:
> I have one question: after ignoring some methods, don't we have some extra untested latte files? Let's say I have actionDefault (ignored) and default.latte, will default.latte be marked as untested?

If nothing else tests that template it will be reported, I think it is expected. You can add that file to ignored errors if you do not want it to be tested at all. 

IDEA: We probably should also support use of PHPStan `excludePaths`  to also exculde latte files from analysis and reporting it as unanalysed.

src/PhpDoc/LattePhpDocResolver.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Efabrica\PHPStanLatte\PhpDoc;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\PhpDoc\PhpDocStringResolver;
+use PHPStan\Reflection\ReflectionProvider;
+
+class LattePhpDocResolver
```
lulco: Add final and strict please

lulco (approved):

src/PhpDoc/LattePhpDocResolver.php
MartinMystikJonas: Isnt there CS sniff for strict types? I think I saw it somewhere. MAybe slevomat coding standard? Maybe we should add it to automate these checks.

MartinMystikJonas: FOund it https://github.com/slevomat/coding-standard/blob/master/doc/type-hints.md#slevomatcodingstandardtypehintsdeclarestricttypes-

lulco: Yes there is some, I just can’t add it to actual standard because we use it in private projects in company. Maybe I’ll add some new “stricter” standard and use it here

MartinMystikJonas:
I added:
- test (with scanning feature for expected errors/collected in files and no need to write exact fixture namespace in expected errors)
- docs
- updated changelog

Idea: Maybe we should create dir docs a split docs into parts. Like configuration, annotations, how it works ane in README keep on installation, quickstart and links to other parts.

lulco (approved):
Yes, I think we can split docs

PR #128, opened 2022-12-09
TypeSerializer for reliable passing types from collectors

MartinMystikJonas:
All type serialization/deserialization between object and array (passed from colelctors) is done by new service TypeSerializer

### DIFF
--- a/src/Type/TypeSerializer.php
+++ b/src/Type/TypeSerializer.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Type;
+
+use PHPStan\PhpDoc\TypeStringResolver;
+use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\ErrorType;
+use PHPStan\Type\Type;
+use PHPStan\Type\VerbosityLevel;
+use Throwable;
+
+class TypeSerializer
+{
+    private TypeStringResolver $typeStringResolver;
+
+    public function __construct(TypeStringResolver $typeStringResolver)
+    {
+        $this->typeStringResolver = $typeStringResolver;
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    public function toArray(Type $type): array
+    {
+        $typeAsArray = [
+            'precise' => $type->describe(VerbosityLevel::precise()),
+            'basic' => $type->describe(VerbosityLevel::typeOnly()),
+        ];
+        try {
+            $typeAsArray['serialized'] = @serialize($type);
+        } catch (Throwable $e) {
+            // ignore
+        }
+        return $typeAsArray;
+    }
+
+    /**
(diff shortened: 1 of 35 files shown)


src/Type/TypeSerializer.php
```diff
@@ -0,0 +1,67 @@
...
+     * @param TypeStringResolver $typeStringResolver
+     */
+    public function __construct(TypeStringResolver $typeStringResolver)
+    {
+        $this->typeStringResolver = $typeStringResolver;
+    }
+
+    /**
+     * @param Type $type
+     * @return array<string, string>
```
lulco, line 24: shouldn't be `array{precise: string, basic: string, serialized?: string}`? And also everywhere where it is used in value object array types

src/Type/TypeSerializer.php
```diff
@@ -0,0 +1,67 @@
...
+use PHPStan\Type\Type;
+use PHPStan\Type\VerbosityLevel;
+use Throwable;
+
+class TypeSerializer
+{
+    private TypeStringResolver $typeStringResolver;
+
+    /**
+     * @param TypeStringResolver $typeStringResolver
```
lulco: this comment has no additional information

src/Type/TypeSerializer.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public function __construct(TypeStringResolver $typeStringResolver)
+    {
+        $this->typeStringResolver = $typeStringResolver;
+    }
+
+    /**
+     * @param Type $type
```
lulco: this parameter comment has no additional information

src/Type/TypeSerializer.php
```diff
@@ -0,0 +1,67 @@
...
+            $typeAsArray['serialized'] = @serialize($type);
+        } catch (Throwable $e) {
+            // ignore
+        }
+        return $typeAsArray;
+    }
+
+    /**
+     * @param array<string, string> $typeAsArray
+     * @return Type
```
lulco: no additional value

src/Type/TypeSerializer.php
MartinMystikJonas: My idea was taht exact format of serialized type is internal information of TypeSerializer - we might change/add keys and it never should be accessed outside this class

MartinMystikJonas: I would either keep it this was or create type alias TypeAsArray that would be used across app? What would you preffer?

MartinMystikJonas:
Should I merge it already or add that TypeAsArray type alias first ?

src/Type/TypeSerializer.php
lulco: Ok if we don’t need it outside of this class we can leave it string, string

lulco (approved):

lulco:
> Should I merge it already or add that TypeAsArray type alias first ?

You can merge

Issue #130, opened 2022-12-09
Unknown type in foreach variable when $iterator is used inside loop


MartinMystikJonas:
In Latte 3 (not sure about Latte 2) latte code:
```latte
{foreach $registrations as $registration}
```
is compiled as:
```php
foreach ($iterator = $ʟ_it = new Latte\Essential\CachingIterator($registrations, $ʟ_it ?? null) as $registration) {
```
but this wrapping interated collection inside `CachingIterator` lost information about item type. So `$registration` is mixed.

MartinMystikJonas:
How can we solve it? First option might be to use nodeVIsitors to replace
```php
foreach ($iterator = $ʟ_it = new Latte\Essential\CachingIterator($registrations, $ʟ_it ?? null) as $registration) {
```
to
```php
$iterator = $ʟ_it = new Latte\Essential\CachingIterator($registrations, $ʟ_it ?? null);
foreach ($registrations  as $registration) {
```
but not sure how difficult it would be.

Another option might be to correctly define stub for CachingIterator so it would propagate type information.

lulco:
I think I've done this before. But I don’t know where :D

lulco:
@MartinMystikJonas what exactly is the problem? 
I checked the types inside foreach and it works:
https://github.com/efabrica-team/phpstan-latte/actions/runs/3661805309/jobs/6190373603

I have list of strings and in foreach it is string.

MartinMystikJonas:
@lulco I  had reports errors in our real world app abou cslling methods on mixed. I thought it was because of foreach but maybe there is another cause. I will investigate more.

lulco:
Maybe as soon as you use $iterator inside foreach? I did this PR on mobile :) I wasn't able to use more complicated constructs

MartinMystikJonas:
I will check later today

MartinMystikJonas:
Seems like error is somewhere else. Closing.

lulco:
It seems I was right, when I used `$iterator` inside foreach, compiled template is different:
```
$iterations = 0;
foreach ($localStrings as $string) {
    \PhpStan\dumpType($string);
    $iterations++;
}
```
vs
```
$iterations = 0;
foreach ($iterator = $ʟ_it = new LR\CachingIterator($localStrings, $ʟ_it ?? null) as $string) {
    echo LR\Filters::escapeHtmlText($iterator->counter);
    \PhpStan\dumpType($string);
    $iterations++;
}
```

[...]

PR #131, opened 2022-12-09
Foreach tests and added transform for foreach with iterator

lulco:
Resolves to https://github.com/efabrica-team/phpstan-latte/issues/130

### DIFF
--- a/src/Compiler/NodeVisitor/TransformForeachWithIteratorNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/TransformForeachWithIteratorNodeVisitor.php
@@ -5,9 +5,11 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor;
 
 use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\ActualClassNodeVisitorBehavior;
+use Efabrica\PHPStanLatte\Resolver\NameResolver\NameResolver;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Name;
 use PhpParser\Node\Stmt\Expression;
 use PhpParser\Node\Stmt\Foreach_;
 use PhpParser\NodeVisitorAbstract;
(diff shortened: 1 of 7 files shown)


MartinMystikJonas:
Maybe we should keep this test to be sure it still works in future latte versions. And extend it if Infound reason why foreach do problems in our project.

lulco:
Yes sure, just have to finish it - add expected errors to test and add some iterators, inherit iterators etc.

lulco:
@MartinMystikJonas try this branch if it fix your case

src/Compiler/NodeVisitor/TransformForeachWithIteratorNodeVisitor.php
```diff
@@ -0,0 +1,67 @@
...
+
+        if (!$node->expr instanceof Assign) {
+            return null;
+        }
+
+        if (!$node->expr->expr instanceof Assign) {
+            return null;
+        }
+
+        if (!$node->expr->expr->expr instanceof New_) {
```
MartinMystikJonas, line 57: Maybe use more specific checks here to avoid false matches. At least check that zhis creates new instance of CachingIterator class.

MartinMystikJonas (changes requested):

MartinMystikJonas:
> @MartinMystikJonas try this branch if it fix your case

Just tesed. Works prefectly.

src/Compiler/NodeVisitor/TransformForeachWithIteratorNodeVisitor.php
lulco: Well it's not that simple without scope :)

MartinMystikJonas: I would simply check just that `New_` node `class` property is `Name` node with `Latte\Essential\CachingIterator` No neeed for anything more advanced

lulco: Ok thats doable

src/Compiler/NodeVisitor/TransformForeachWithIteratorNodeVisitor.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (!$node instanceof Foreach_) {
+            return null;
+        }
+
+        if (!$node->expr instanceof Assign) {
+            return null;
+        }
+
+        if (!$node->expr->expr instanceof Assign) {
```
MartinMystikJonas, line 53: Maybe use while loop here to find first node that is not `Assign` expression? I am not sure there are/will be always exactly two of them.

lulco: That depends on latte compiler. Now there are two of them always

MartinMystikJonas: I would preffer more robust solution but I think it can wait if issue with different count occur

PR #132, opened 2022-12-10
Move nikic/php-parser to require-dev

spaze:
Hi, it seems nikic/php-parser is not required for normal operation. only for running tests etc. I think it should be moved.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,6 @@
         "php": ">=7.4 <=8.2",
         "ext-json": "*",
         "phpstan/phpstan": "^1.9.2",
-        "nikic/php-parser": "^4.15",
         "latte/latte": "^2.11.6 | ^3.0.4",
         "nette/utils": "^3.2",
         "nette/finder": "^2.5"
@@ -14,6 +13,7 @@
         "phpunit/phpunit": "^9.5",
         "nette/application": "^3.1.2",
         "nette/forms": "^3.1",
+        "nikic/php-parser": "^4.15",
         "symfony/finder": "^5.4 | ^6.0",
         "efabrica/coding-standard": "^0.4"
     },


lulco:
What is "normal" operation? :) We use nodes from php parser in collectors and visitors so without it package will not work

MartinMystikJonas:
What issue do you have with it being in require? Some composer conflict?

spaze:
> What is "normal" operation? :)

Running `phpstan analyze` with the extension loaded.

> We use nodes from php parser in collectors and visitors so without it package will not work

What do you mean by "will not work"? :-)

I have deleted the `vendor/nikic/php-parser` dir right after installing the package and got no errors when running analyze, besides issues with my Latte configuration (unknown tags etc.) of course :-)

I also use parser's nodes in [my extensions](https://github.com/spaze/phpstan-disallowed-calls/blob/8648e5615494833105886e23c3051530e959a6a1/src/Usages/VariableUsages.php#L7) without [having the parser in `require`](https://github.com/spaze/phpstan-disallowed-calls/blob/8648e5615494833105886e23c3051530e959a6a1/composer.json#L23-L34) and they work as expected.

I haven't studied the Latte extension in detail but it seems that the parser lib provided by PHPStan itself should be sufficient.

spaze:
> What issue do you have with it being in require? Some composer conflict?

I just don't like unused code in my repository :-)

lulco:
I was trying to make sure that this package will be installed with php parser of some version (because in older versions was something different). With package in require-dev, user can have installed those older versions and my defined version is ignored, so I am not sure if it will work correctly

MartinMystikJonas:
🤔 What kinf of magic is this? 😁

Well maybe PHPStan requirement is enough when if it bundles PHPParser correctly.

lulco:
I have to check it, I think in prefer lowest tests there was some problem. Maybe we could use require dev + conflict?

spaze:
I have installed nikic/php-parser 4.12.0 which doesn't support `CallLike` nodes ([added in 4.13.0](https://github.com/nikic/PHP-Parser/blob/4.x/CHANGELOG.md#version-4130-2021-09-20)) and my extension [uses it](https://github.com/spaze/phpstan-disallowed-calls/blob/cb8db9cb157fda36091b790e9d9399a5e649e707/src/DisallowedHelper.php#L37) and unless I've done something horribly wrong, the bundled version was used (tests have finished green) instead of the [manually installed one](https://github.com/spaze/michalspacek.cz/actions/runs/3664942421/jobs/6195689936#step:6:13).

spaze:
Heck, I even installed 1.0.0 🤣 https://github.com/spaze/michalspacek.cz/actions/runs/3664981578/jobs/6195756568#step:6:13
and the tests are still green https://github.com/spaze/michalspacek.cz/actions/runs/3664981578/jobs/6195756568#step:7:9

spaze:
👍🙏

PR #133, opened 2022-12-10
Correct Latte3 extensions config key and value

spaze:
When trying to set this up with Latte 3, I've noticed the doc config key isn't correct. Eventually went with `engineBootstrap` config but here's what I found. Hope it's useful at least a bit.


MartinMystikJonas:
Thank you!

PR #135, opened 2022-12-10
Prevent uninitalized property error when no templates was analysed in current run

MartinMystikJonas:


lulco (approved):

lulco:
We can turn on phpstan rule for it.

MartinMystikJonas:
> We can turn on phpstan rule for it.

Not sure what do you mean

lulco:
```neon
parameters:
    checkUninitializedProperties: true
```

lulco:
https://phpstan.org/config-reference#checkuninitializedproperties

MartinMystikJonas:
Oh yes I think we should turn it on.

PR #139, opened 2022-12-11
Transform Block methods and Block calls

### LINKED ISSUE #136: Block parameters have always unknown type
When block with parameters is use type of parameters is always unknown even when type is specified in latte.

Latte template:
```latte
{define statistic-row \App\Model\Dto\StatisticsRow $row, ?bool $negative, ?bool $emphasize}
  {* ... *}
{/define}
{* ... *}
{include statistic-row 'row' => $statistic, 'negative' => true}
```

is compiled as:
```php
    /** {define statistic-row \App\Model\Dto\StatisticsRow $row, $negative, $emphasize} on line 1 */
    public function blockStatistic_row(array $ʟ_args) : void
    {
        extract($this->params);
        // [only global variable types]
        $row = $ʟ_args[0] ?? $ʟ_args['row'] ?? null; // <-- $row type is mixed
        $negative = $ʟ_args[1] ?? $ʟ_args['negative'] ?? null;
        $emphasize = $ʟ_args[2] ?? $ʟ_args['emphasize'] ?? null;
        // ...
    }

    //...
    $this->renderBlock('statistic-row', ['row' => $statistic] + [], 'html');
    // ...
```

### LINKED ISSUE #137: Evaluate parameters mismatch for blocks
Call of block with params is currently evaluated as
```php
$this->renderBlock('statistic-row', ['row' => $statistic] + [], 'html');
```

so it does not check if call is done with right parameter types.

Compiled block definition looks like this:
```php
    /** {define statistic-row \App\Model\Dto\StatisticsRow StatisticsRow $row, bool $negative, bool $emphasize} on line 1 */
    public function blockStatistic_row(array $ʟ_args) : void
```

Not much chance to statical analysis of that too.

Maybe for analysis purposes we can transform block signatures based on doc comment to:
```php
public function blockStatistic_row(\App\Model\Dto\StatisticsRow StatisticsRow $row, bool $negative, bool $emphasize) : void
```
and replace block call with:
```php
$this->blockStatistic_row($statistics);
```

But there could be some magic with blocks from other templates etc. It would need better examination.

MartinMystikJonas:
Resolves #136 
Resolves https://github.com/efabrica-team/phpstan-latte/issues/137


MartinMystikJonas:
```php
    /** {define my-block, \stdClass $paramObject, string $paramString, ?string $paramNullable, $paramNoType, string $paramDefault = 'default', $paramNoTypeDefault = 'default'} on line 1 */
    public function blockMy_block(array $ʟ_args) : void
    {
        extract($this->params);
        /** GLOBAL VARIABLES */
        $paramObject = $ʟ_args[0] ?? $ʟ_args['paramObject'] ?? null;
        $paramString = $ʟ_args[1] ?? $ʟ_args['paramString'] ?? null;
        $paramNullable = $ʟ_args[2] ?? $ʟ_args['paramNullable'] ?? null;
        $paramNoType = $ʟ_args[3] ?? $ʟ_args['paramNoType'] ?? null;
        $paramDefault = $ʟ_args[4] ?? $ʟ_args['paramDefault'] ?? 'default';
        $paramNoTypeDefault = $ʟ_args[5] ?? $ʟ_args['paramNoTypeDefault'] ?? 'default';
        unset($ʟ_args);
```

should be:

```php
    /**
    * {define my-block, \stdClass $paramObject, string $paramString, ?string $paramNullable, $paramNoType, string $paramDefault = 'default', $paramNoTypeDefault = 'default'} on line 1
    * @param ?\stdClass $paramObject Always nullable - all params are optional in latte
    * @param ?string $paramString
    * @param ?string $paramNullable
    * @param $paramNoType
    * @param ?string $paramDefault
    * @param ?string $paramNoTypeDefault
    */
    public function blockMy_block(array $ʟ_args, $paramObject = null, $paramString = null, $paramNullable = null, $paramNoType = null, $paramDefault = null, $paramNoTypeDefault = null) : void
    {
        extract($this->params);
        /** GLOBAL VARIABLES */
        $paramObject = $paramObject ?? null; 
        $paramString = $paramString ?? null;
        $paramNullable = $paramNullable ?? null;
        $paramNoType = $paramNoType ?? null;
        $paramDefault = $paramDefault ?? 'default'; // <== define default here or in method signature what is easier
        $paramNoTypeDefault = $paramNoTypeDefault ?? 'default';
        unset($ʟ_args);
```

src/Compiler/NodeVisitor/AddParametersForBlockNodeVisitor.php
```diff
@@ -0,0 +1,108 @@
...
+            $nodeComment .= "\n/**\n";
+            for ($i = 0; $i < count($parameters[0]); $i++) {
+                // Type is always nullable - all params are optional in latte
+                $type = ltrim(trim($parameters['type'][$i]), '?');
+                $nodeComment .= ' * @param ' . ($type ? '?' . $type . ' ' : '') . '$' . $parameters['variable'][$i] . "\n";
+
+                /** @var string|null $defaultValue */
+                $defaultValue = $parameters['default'][$i] ?: null;
+                if ($defaultValue !== null && str_starts_with('\'', $defaultValue)) {
+                    $default = new String_(str_replace('\'', '', $defaultValue));
```
MartinMystikJonas: this could replace ' that is valid part of the string content. Probably removing first and last character would be better

lulco: True

lulco:
Self note: In add var types skip those lines which are defined as method parameters in block* methods

tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Blocks/define.latte
```diff
@@ -0,0 +1,13 @@
...
+    {php \PhpStan\dumpType($paramNullable)}
+    {php \PhpStan\dumpType($paramNoType)}
+    {php \PhpStan\dumpType($paramDefault)}
+    {php \PhpStan\dumpType($paramNoTypeDefault)}
+    {php \PhpStan\dumpType($paramFloat)}
+    {php \PhpStan\dumpType($paramInteger)}
+{/define}
+
+{include my-block 'paramObject' => $knownObject, 'paramString' => $knownString, 'paramNullable' => null, 'paramNoType' => $knownString, 'paramFloat' => $knownFloat, 'paramInteger' => $knownInteger}
+{include my-block 'paramObject' => $knownString, 'paramString' => $knownInteger}
```
lulco, line 13: @MartinMystikJonas is it possible to call it without keys of parameters? Only include $var1, $var2 …

tests/Rule/LatteTemplatesRule/SimpleControl/LatteTemplatesRuleForSimpleControlTest.php
```diff
@@ -196,4 +196,50 @@ public function testResolve(): void
...
+            [
+                'Dumped type: float|null',
+                8,
+                'define.latte',
+            ],
+            [
+                'Dumped type: int|null',
+                9,
+                'define.latte',
+            ],
```
lulco, line 252: Todo add test cases for wrong calls

tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Blocks/define.latte
MartinMystikJonas: Yes it is

src/Compiler/NodeVisitor/AddParametersForBlockNodeVisitor.php
```diff
@@ -0,0 +1,108 @@
...
+                $node->params[] = new Param(new Variable($parameters['variable'][$i]), $default);
+            }
+            $nodeComment .= '*/';
+        }
+        $node->setDocComment(new Doc($nodeComment));
+
+        $stmts = (array)$node->stmts;
+        $newStmts = [];
+        foreach ($stmts as $stmt) {
+            if ($stmt instanceof Expression && $stmt->expr instanceof Assign) {
```
lulco: Check if it assign from L_arg

lulco: Remove also unset(L_arg)

src/Compiler/NodeVisitor/RenderBlockNodeVisitor.php
```diff
@@ -0,0 +1,137 @@
...
+        if (!$paramsArg->value instanceof Plus) {
+            return null;
+        }
+
+        if (!$paramsArg->value->left instanceof Array_) {
+            return null;
+        }
+
+        $params = [
+            'ʟ_args' => new Array_(),
```
lulco: Try to completelly remove L_args

src/Compiler/NodeVisitor/RenderBlockNodeVisitor.php
```diff
@@ -0,0 +1,137 @@
...
+        foreach ($blockMethodParams as $blockMethodParam) {
+            if (!$blockMethodParam instanceof Param) {
+                continue;
+            }
+            if (!$blockMethodParam->var instanceof Variable) {
+                continue;
+            }
+            $variableName = $this->nameResolver->resolve($blockMethodParam->var->name);
+
+//            var_dump($variableName);
```
lulco: Remove debug

tests/Rule/LatteTemplatesRule/SimpleControl/LatteTemplatesRuleForSimpleControlTest.php
lulco: added, also added errors about undefined types

tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Blocks/define.latte
lulco: OK I think I solved it

MartinMystikJonas:
Looks good to me. I will try it on real app now.

lulco:
@MartinMystikJonas I thing is missing - ignore variables from parameters, it needs NodeVisitors refactoring, I created issue. I have to analyse which Visitors can run together and what parameters they need (not all have only parameters from constructor).

MartinMystikJonas:
Yes the error persists and it seems wrong order of node visitors is the cause.

lulco:
Yes, but when I switch them I got another group of errors :) I have to analyse it first - here is the issue https://github.com/efabrica-team/phpstan-latte/issues/144

But the main task is done, so I would merge it. Imho it is very bad practice to name "global" and "local" variables with the same name

MartinMystikJonas:
Ok I think we could merge it for now. But I would welcome to solve it soon because it causes errors in my code.

Our code is a bit specific in this because it uses some magic with blocks for datagrid rendering and I need to have variable with same name as global for some blocks and for some I can define it as parameter.

PR #140, opened 2022-12-11
Dynamic include support and advanced expression resolving for variable values

### LINKED ISSUE #141: Dynamic includes cannot ve evaluated
Test case ready in #140

MartinMystikJonas:
Test case for dynamic includes (included path passed in variable).

Not sure how this should be solved. It seems that root cause is that variable type is not passed as constant string but as `literal-string&non-falsy-string`. Maybe just use ValueResolver on variable values in VariableCollector?

---

Implemented. Resolved #141

### DIFF
--- a/src/Collector/VariableCollector.php
+++ b/src/Collector/VariableCollector.php
@@ -7,6 +7,7 @@ namespace Efabrica\PHPStanLatte\Collector;
 use Efabrica\PHPStanLatte\Collector\ValueObject\CollectedVariable;
 use Efabrica\PHPStanLatte\PhpDoc\LattePhpDocResolver;
 use Efabrica\PHPStanLatte\Resolver\TypeResolver\TemplateTypeResolver;
+use Efabrica\PHPStanLatte\Resolver\TypeResolver\TypeResolver;
 use Efabrica\PHPStanLatte\Template\Variable as TemplateVariable;
 use Efabrica\PHPStanLatte\Type\TypeSerializer;
 use PhpParser\Node;
@@ -22,13 +23,16 @@ use PHPStan\Analyser\Scope;
  */
 final class VariableCollector extends AbstractCollector
 {
+    private TypeResolver $typeResolver;
+
     private TemplateTypeResolver $templateTypeResolver;
 
     private LattePhpDocResolver $lattePhpDocResolver;
 
-    public function __construct(TypeSerializer $typeSerializer, TemplateTypeResolver $templateTypeResolver, LattePhpDocResolver $lattePhpDocResolver)
+    public function __construct(TypeSerializer $typeSerializer, TypeResolver $typeResolver, TemplateTypeResolver $templateTypeResolver, LattePhpDocResolver $lattePhpDocResolver)
     {
         parent::__construct($typeSerializer);
+        $this->typeResolver = $typeResolver;
         $this->templateTypeResolver = $templateTypeResolver;
         $this->lattePhpDocResolver = $lattePhpDocResolver;
     }
@@ -82,11 +86,16 @@ final class VariableCollector extends AbstractCollector
             return null;
         }
 
+        $variableType = $this->typeResolver->resolveAsConstantType($node->expr, $scope);
+        if ($variableType === null) {
+            $variableType = $scope->getType($node->expr);
+        }
+
         $variableName = is_string($nameNode) ? $nameNode : $nameNode->name;
         return $this->collectItem(new CollectedVariable(
             $classReflection->getName(),
             $functionName,
-            new TemplateVariable($variableName, $scope->getType($node->expr))
+            new TemplateVariable($variableName, $variableType)
         ));
     }
 }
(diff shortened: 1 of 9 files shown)


lulco:
Maybe we can use it for string type variables to precise type?

lulco:
We have to try if it will help

MartinMystikJonas:
I am already experimenting with it

lulco:
I'm little bit afraid of performance issues but me personally don’t care much about few extra seconds if it will solve some problem

MartinMystikJonas:
Implemented and it is working. I do not think performance will be impacted significantly.

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/VariablesPresenter.php
```diff
@@ -48,4 +48,9 @@ public function actionDirectRender(): void
         $this->template->render(__DIR__ . '/templates/Variables/direct.latte', ['fromRender' => 'b']);
         $this->terminate();
     }
+
+    public function actionDynamicInclude(): void
+    {
```
lulco, line 53: Maybe we should add some variable which is successfully passed to included template

src/Resolver/ValueResolver/ValueResolver.php
```diff
@@ -104,4 +110,29 @@ public function resolve(Expr $expr, Scope $scope, $unknownValuePlaceholder = nul
             return null;
         }
     }
+
+    public function resolveAsConstantType(Expr $expr, Scope $scope): ?Type
```
lulco: Is this reallly part of ValueResolver? Imho you’re returning Type, so it should be method in TypeResolver.

src/Resolver/ValueResolver/ValueResolver.php
```diff
@@ -104,4 +110,29 @@ public function resolve(Expr $expr, Scope $scope, $unknownValuePlaceholder = nul
...
+                $types[] = new ConstantStringType($value);
+            } elseif (is_int($value)) {
+                $types[] = new ConstantIntegerType($value);
+            } elseif (is_float($value)) {
+                $types[] = new ConstantFloatType($value);
+            } elseif (is_bool($value)) {
+                $types[] = new ConstantBooleanType($value);
+            } elseif (is_null($value)) {
+                $types[] = new NullType();
+            } else {
```
lulco: So if one of 5 values can’t be covered we silently return null? Imho we should return union type of 4/5 precisio or throw exception

src/Collector/VariableCollector.php
```diff
@@ -82,11 +86,16 @@ public function processNode(Node $node, Scope $scope): ?array
             return null;
         }
 
+        $variableType = $this->valueResolver->resolveAsConstantType($node->expr, $scope);
+        if ($variableType === null) {
```
lulco, line 90: This can be directly done in TypeResolver, you send scope there

src/Resolver/ValueResolver/ValueResolver.php
lulco: Or get type by scope :)

lulco (changes requested):

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/VariablesPresenter.php
MartinMystikJonas: Ok

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: For me it seems closely related to vakue resolver. It is just different representation or resolved value.

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: We use null as "cannot be resolved" in value resolvers. I would rather avoid exception for non-exceptional result of resolving.

src/Collector/VariableCollector.php
MartinMystikJonas: It could but fallback to generic type from scope seems to me like VariableCollector specific behaviour that probably woukd not be wanted in other potential uses (there are none currently but I think they might be)

lulco: Ok if we will create another use of it, we will see if scope fallback will be there too

src/Resolver/ValueResolver/ValueResolver.php
lulco: Yes thats why I wrote to get type from scope

src/Resolver/ValueResolver/ValueResolver.php
lulco: Well closely it is in this case but you just resolving Type I would like to split it

lulco (changes requested):

src/Resolver/ValueResolver/ValueResolver.php
MartinMystikJonas: So create ConstantValueTypeResolver in TypeaResolvers that would have ValueResolver as dependency?

lulco: I prefer TypeResolver with method resolveAsConstantType

MartinMystikJonas:
Updated. Should be ready to merge.

lulco (approved):

lulco:
thank you!

PR #146, opened 2022-12-13
Annotation @phpstan-latte-template

MartinMystikJonas:


MartinMystikJonas:
Just draft for now. I need to add docs and maybe some more tests

MartinMystikJonas:
Ready for review

lulco:
do we need all those empty files `{something}.php.latte`?

MartinMystikJonas:
I am testing if the are resolved/checked based on annotation. I left them emptxy because absence of missing tempalte file... error is enough for my tests

docs/annotations.md
```diff
@@ -96,3 +97,89 @@ class MyControl extends Control
     }
 }
 ```
+
+## `@phpstan-latte-template`
+Allowed context: render call, method, class
+
+Value specifies what template or templates should be resolved and checked in given context.
+
+Multiple annotations can be used on single element or at different levels to check multiple tempaltes with same context.
```
lulco: you have always typo in word templates :) but only in discussions / readme, in code it is usually correct

docs/annotations.md
```diff
@@ -96,3 +97,89 @@ class MyControl extends Control
...
+
+### Behaviour
+
+When this annotation is used on render call it will replace value passed as tempalte to render call.
```
lulco: tempalte again

docs/annotations.md
```diff
@@ -96,3 +97,89 @@ class MyControl extends Control
...
+
+When annotation is used on method it will resolve given template in context of this method (alongside renders used in method body).
+
+When annotation is used on class it ats like it was added to all methods in class.
```
lulco: acts?

docs/annotations.md
```diff
@@ -96,3 +97,89 @@ class MyControl extends Control
...
+
+### Annotation value
+
+Can use wildcard `*` to match multiple files.
```
lulco: You can? or Wildcard can be used?

docs/annotations.md
```diff
@@ -96,3 +97,89 @@ class MyControl extends Control
...
+
+Can use placeholders:
+- `{dir}` = `__DIR__` (example `/app/dir/controls`)
+- `{file}` = `__FILE__` (example `/app/dir/controls/MyControl.php`)
+- `{fileName}` = `basename(__FILE__)` (example `MyControl.php`)
+- `{fileNameNoSuffix}` = file anme with suffix removed (example `MyControl`)
```
lulco: we could naming like https://www.php.net/manual/en/function.pathinfo.php
filename, basename, extension

docs/annotations.md
```diff
@@ -96,3 +97,89 @@ class MyControl extends Control
...
+- `{classShortName}` = class short name withnout namespace (example `MyControl`)
```
lulco: I think we can call it className, and for FQN we can use fullClassName or so

lulco (approved):
lgtm

docs/annotations.md
MartinMystikJonas: Yes I have few such typos in muscle memory and cannot get rid of them

PR #147, opened 2022-12-13
Fixed error when analysing path without any latte templates

MartinMystikJonas:


lulco:
wasn't this fixed already?

lulco (approved):

MartinMystikJonas:
Was fixed for no analysed templates. But when there are no templates at all it still caused this.

PR #148, opened 2022-12-13
Fixed error when analysing classes using self:: or static:: in method calls

### LINKED ISSUE #145: Class self was not found while trying to analyse it - discovering symbols is probably not configured properly.
We got this error in real app when extension is active. It is not caused by rule becaus eerror occurs even without it. My guess it is somewhere in collectors but I need to investigate.

MartinMystikJonas:
Resolves #145

### DIFF
--- a/src/Resolver/CallResolver/CalledClassResolver.php
+++ b/src/Resolver/CallResolver/CalledClassResolver.php
@@ -37,7 +37,7 @@ final class CalledClassResolver
                     return null;
                 }
                 return $classReflection->getName();
-            } elseif($calledClassName === 'self' || $calledClassName === 'static')  {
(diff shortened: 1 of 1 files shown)


lulco (approved):

src/Resolver/CallResolver/CalledClassResolver.php
```diff
@@ -37,6 +37,8 @@ public function resolve(Node $node, Scope $scope): ?string
                     return null;
                 }
                 return $classReflection->getName();
+            } elseif($calledClassName === 'self' || $calledClassName === 'static')  {
```
lulco: one space is missing here, but Checks don't run now, I don't know why

lulco:
This is approved

PR #152, opened 2022-12-14
Traits

### LINKED ISSUE #79: Collect data from used traits in CollectorFinders
variables, components and / or called methods can be defined also in traits used by Presenters/Controls

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/79


lulco:
@MartinMystikJonas I wasn't able to use NodeScopeResolver nor Analyser, so I've wrote own simple trait processing in component collector and it works. I will add it to AbstractCollector and use also in other collectors.

src/Collector/ComponentCollector.php
```diff
@@ -73,6 +109,25 @@ public function processNode(Node $node, Scope $scope): ?array
...
 
+    /**
+     * @param Node[] $stmts
+     * @phpstan-return CollectedComponentArray[]
+     */
+    private function processStmts(array $stmts, Scope $scope): array
+    {
+        $return = [];
+        foreach ($stmts as $stmt) {
+            if (isset($stmt->stmts)) {
```
lulco: use getSubNodeNames() and foreach

src/Collector/MethodCallCollector.php
```diff
@@ -109,28 +109,11 @@ public function processNode(Node $node, Scope $scope): ?array
         }
 
         $declaringClassName = $reflectionMethod->getDeclaringClass()->getName();
-
-        // Do not find template variables in external classes
-        if ($this->isExternalCall($declaringClassName, $calledMethodName)) {
```
MartinMystikJonas, line 114: Why is this removed?

src/Collector/ComponentCollector.php
```diff
@@ -51,6 +66,27 @@ public function getNodeType(): string
...
+                }
+
+                $traitClassReflection = $this->reflectionProvider->getClass($traitName);
+                $traitFileName = $traitClassReflection->getFileName();
+                if ($traitFileName === null) {
+                    continue;
+                }
+
+                $traitNodes = $this->parser->parseFile($traitFileName);
+                $traitComponents = array_merge($traitComponents, $this->processStmts($traitNodes, $scope));
```
MartinMystikJonas: Are you sure this will not corrupt Scope with unvanted changes? Mybe clone Scope before passing it?

src/Collector/ComponentCollector.php
```diff
@@ -73,6 +109,25 @@ public function processNode(Node $node, Scope $scope): ?array
...
+     * @phpstan-return CollectedComponentArray[]
+     */
+    private function processStmts(array $stmts, Scope $scope): array
+    {
+        $return = [];
+        foreach ($stmts as $stmt) {
+            if (isset($stmt->stmts)) {
+                $return = array_merge($return, $this->processStmts($stmt->stmts, $scope));
+            }
+            $procesNodeReturn = $this->processNode($stmt, $scope);
```
MartinMystikJonas: Are you sure Scope will have correct data for variable types etc? At first sight it seems you bypassed PHPStan logic that allows expression type resolving by Scope. You might have scope freezed at point of trait use used to analyse all content of trait.

MartinMystikJonas:
I am not sure it will work as intended when trait content would need anything more complex. Would this approach work for VAriableCollector for example where we need Scope to resolve variable type at point of assign node?

src/Collector/ComponentCollector.php
lulco: I believe it will work.
1) if I have scope for trait, I don’t have class reflection
2) traits are copied to classes in runtime
3) I think I saw it in phpstan itself to use classScope on traits

lulco: Also I believe someone will figure out how to use phpstan built-in methods. I just don’t want to freeze at this point so long :)

src/Collector/MethodCallCollector.php
lulco: I think it is unreachable because of collector paths added in previous commits

src/Collector/ComponentCollector.php
MartinMystikJonas: Problem is that you have frozen scope.  You are not updating it based on analyzed statements. In PHPStan Scope changes after every analysed statement. For example after variable assign Scope remembers what type was assigned to variable so later when you ask scope to getType of that variable it is known. This will not work in your setup I am afraid. Only variables known at the start will be known. Nothing that happens inside trait will update scope.

MartinMystikJonas: This code in Trait will failed to resolve I am afraid:
```php
$var = 'string'; // <-- analysis of this statement will not update Scope
$this->emplate->myVariable = $var; // <-- VariableCollector would resolve myVariable as mixed because information that $var exists and is string will not be updated in Scope during trait analysis
```

lulco: I see, and how to change scope? :) it's black magic for me. I was so excited it works… you calm down me :D

MartinMystikJonas: Well that is what NodeScopeResolver do as far as I understand

lulco: well I tried it and it can find types of variables, but it don't know in which method it is assigned...

MartinMystikJonas: Hmm I wonder how information of variable type got into scope. What code you tested?

src/Collector/ComponentCollector.php
lulco: scope changed

lulco:
I spend few hours with this and then I figure out that I had variables from trait collected, but they were collected for PresenterClass, but Finder looks into declaring class and it is Trait... so I changed the collector for variables and now it works. Almost... :) Like you say, now assign of type:
```
$var = 'string'; // <-- analysis of this statement will not update Scope
$this->emplate->myVariable = $var; // <-- VariableCollector would resolve myVariable as mixed because information that $var exists and is string will not be updated in Scope during trait analysis
```
not working. It's just about some changing scope, but there are so many steps in NodeScopeResolver I wouldn't reproduce.

I would like to merge this, as it solves my actual problems. And maybe if you will have time to try somehow use / setup NodeScopeResolver before `FileAnalyser->analyseFile()`, I would appreciate, because I think I tried every combination of calls, and that just not working. Maybe I'm blind in this.

src/Collector/ComponentCollector.php
lulco: I'm not sure now, maybe I hadn't this type of assign that time. You're right this is not working now.

MartinMystikJonas:
Ok Lets keep this workaround if it solves your problem and I will try to play with it and find better sokution using NodeScopeResolver later

lulco:
Thanks, I just got idea that maybe traits were analysed with my tries, but they were in other scope so just didn’t passed throght some condition in collectors - similar to Variables where I had to switch class reflection to trait one.

Issue #143, opened 2022-12-12
Custom Compiler/Engine for some templates


MartinMystikJonas:
When Latte is used standalone for example to generate e-mail content we want to have option to say that given latte template is compiled using different Compiler/Engine than default.

All classes that currently get Compiler (CompilerInterface) as service from DI should accept it as parameter and use different compiler.

LatteTemplateResolver should be able to say that resolved templates use different Compiler/Engine. Maybe make it part od Template Value object.

Issue #58, opened 2022-11-27
Collect filters added in component/presenter


MartinMystikJonas:
Components and presenters can add custom filters to templates. It should be collected and passed to template.

PR #155, opened 2022-12-16
Refactoring of LatteToPhpCompiler

lulco:
Refactoring of LatteToPhpCompiler with the prospect of further changes:
- filters (https://github.com/efabrica-team/phpstan-latte/issues/58)
- engine (https://github.com/efabrica-team/phpstan-latte/issues/143)

### DIFF
--- a/src/Compiler/LatteToPhpCompiler.php
+++ b/src/Compiler/LatteToPhpCompiler.php
@@ -4,10 +4,8 @@ declare(strict_types=1);
 
 namespace Efabrica\PHPStanLatte\Compiler;
 
-use Efabrica\PHPStanLatte\Collector\ValueObject\CollectedForm;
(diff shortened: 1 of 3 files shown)


MartinMystikJonas:
I was thinking about similar change. But maybe pass Template to all visitors instead of different setup for each?

MartinMystikJonas:
Maybe remove ActualClassNodeVisitorInterface and make all visitors implement setTemplate(). All visitors would have same setup in DI and all of then would be initialized same way with setTemplate() - and each of them can use what it needs from Template

lulco:
Maybe later, now there is at least AddVarTypesNodeVisitor which don't read variables always from Template. I want to create several interfaces like ActualClassNodeVisitorInteraface for Forms, Components, Filters etc. and each visitor will get only what it needs.

MartinMystikJonas:
ok

MartinMystikJonas (approved):

PR #160, opened 2022-12-16
Annotation @phpstan-latte-var

MartinMystikJonas:


src/Template/Variable.php
```diff
@@ -44,4 +45,43 @@ public function jsonSerialize()
           'type' => $this->getTypeAsString(),
         ];
     }
+
+    /**
+     * @param array<Variable> ...$variableArrays
+     * @return array<Variable>
+     */
+    public static function union(array ...$variableArrays): array
```
MartinMystikJonas: This logic was used on multiple places so i extracted it here

src/Collector/VariableCollector.php
```diff
@@ -74,28 +73,38 @@ public function processNode(Node $node, Scope $scope): ?array
         }
 
         if ($nameNode instanceof Expr) {
-            return null;
+            $variableName = null;
```
MartinMystikJonas, line 88: I need to catch assignments that cannot be evaluted but can have annotation for example `$this->template->{$varName}` can be annotated to allow analysis

docs/annotations.md
```diff
@@ -20,7 +24,7 @@ class MyControl extends Control
             /** @phpstan-latte-ignore */
             $this->template->something = $param; // <-- this variable will not be collected (by default all template variables assigned inside method are collected)
         } else {
-            $this->render(__DIR__ . '/MyControl.latte');
+            $this->template->render(__DIR__ . '/MyControl.latte');
```
MartinMystikJonas, line 27: Example code was invalid I fixed it

MartinMystikJonas:
I need to fix tests. It works on local but have some errors on CI.

MartinMystikJonas:
@lulco Tests fixed ready for review

src/Collector/Finder/VariableFinder.php
```diff
@@ -44,11 +54,10 @@ public function __construct(CollectedDataNode $collectedDataNode, TypeSerializer
      */
     public function find(string $className, string $methodName): array
     {
-        return array_merge(
-            $this->collectedVariables[$className][''] ?? [],
-            $this->findInParents($className),
+        return Variable::merge(
```
lulco: This is confusing for me. Why Variable has static merge method where all other variables are somehow merged? I would create some service with non static method or at least helper.

src/Collector/Finder/VariableFinder.php
```diff
@@ -18,24 +19,33 @@
...
-            $methodName = $collectedVariable->getMethodName();
-            if (!isset($this->collectedVariables[$className][$methodName])) {
-                $this->collectedVariables[$className][$methodName] = [];
-            }
-            $this->collectedVariables[$className][$methodName][] = $collectedVariable->getVariable();
+        $assignedVariables = VariableCollector::loadData($collectedDataNode, $typeSerializer, CollectedVariable::class);
+        foreach ($assignedVariables as $variable) {
+            $className = $variable->getClassName();
+            $methodName = $variable->getMethodName();
+            $this->assignedVariables[$className][$methodName] = Variable::union($this->assignedVariables[$className][$methodName] ?? [], [$variable->getVariable()]);
```
lulco: Union on variable should just union actual type of variable with new type. Not create new array of all variables. This should be service / helper outside of Variable

lulco (changes requested):
This PR is huge again, I need bigger display.

src/Collector/Finder/VariableFinder.php
MartinMystikJonas: I was thinking about helper too. I will change it.

MartinMystikJonas:
Well most of it are tests :-) There are huge amount of combinations taht needed to be tested.

src/Collector/Finder/VariableFinder.php
MartinMystikJonas: Any idea of good name? Simply VariableHelper?

lulco: probably yes :)

MartinMystikJonas:
Updated and rebased on main

PR #161, opened 2022-12-17
Collect filters

### LINKED ISSUE #58: Collect filters added in component/presenter

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/58

### DIFF
--- a/src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor;
 
 use Closure;
-use Efabrica\PHPStanLatte\Compiler\Compiler\CompilerInterface;
 use Efabrica\PHPStanLatte\Compiler\LatteVersion;
 use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\ActualClassNodeVisitorBehavior;
 use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\ActualClassNodeVisitorInterface;
@@ -31,9 +30,12 @@ final class ChangeFiltersNodeVisitor extends NodeVisitorAbstract implements Actu
     /** @var array<string, string|array{string, string}|array{object, string}|callable> */
     private array $filters;
 
-    public function __construct(CompilerInterface $compiler)
+    /**
+     * @param array<string, string|array{string, string}|array{object, string}|callable> $filters
+     */
+    public function __construct(array $filters)
     {
-        $this->filters = $compiler->getFilters();
+        $this->filters = LatteVersion::isLatte2() ? array_change_key_case($filters) : $filters;
     }
 
     public function enterNode(Node $node): ?Node
--- a/src/VariableCollector/DynamicFilterVariables.php
+++ b/src/VariableCollector/DynamicFilterVariables.php
@@ -11,6 +11,7 @@ use Efabrica\PHPStanLatte\Template\Variable;
 use PHPStan\BetterReflection\BetterReflection;
 use PHPStan\Broker\ClassNotFoundException;
 use PHPStan\PhpDoc\TypeStringResolver;
+use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\CallableType;
 use PHPStan\Type\ClosureTypeFactory;
 use PHPStan\Type\ObjectType;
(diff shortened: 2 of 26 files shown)


lulco:
@MartinMystikJonas I have to admit that I sometimes don't know what I'm doing in those Resolver classes / traits, just add some parameter to methods, methods, some calls to globalSomething etc. I'm not sure if it is good pattern there. I had to do it with forms, now with filters, and probably with engines it will be also needed.

MartinMystikJonas:
Maybe instead of traits we should use some utility class to hold logic for presenter/components global data? Something like PresenterContextAnalyser? With methods getGlobalVariables,... 

I am also thinking we need some kind of value object TemplateContext and related finder that would hold variables, components, forms and now filters so adding new context type would not spread across code base like this.

lulco:
We can refactor it later, this PR is big enough :)

MartinMystikJonas:
Agreed. It was just some ideas. I will review it later today. But seems fine at first sight to me.

src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
```diff
@@ -31,9 +30,12 @@ final class ChangeFiltersNodeVisitor extends NodeVisitorAbstract implements Actu
...
     private array $filters;
 
-    public function __construct(CompilerInterface $compiler)
+    /**
+     * @param array<string, string|array{string, string}|array{object, string}|callable> $filters
+     */
+    public function __construct(array $filters)
     {
-        $this->filters = $compiler->getFilters();
+        $this->filters = LatteVersion::isLatte2() ? array_change_key_case($filters) : $filters;
```
MartinMystikJonas, line 38: Isn't this already covered on caller level?

src/VariableCollector/DynamicFilterVariables.php
```diff
@@ -31,13 +35,22 @@ public function __construct(CompilerInterface $compiler, TypeStringResolver  $ty
         $this->closureTypeFactory = $closureTypeFactory;
     }
 
+    /**
+     * @param array<string, string|array{string, string}|array{object, string}|callable> $filters
+     */
+    public function addFilters(array $filters): void
+    {
+        $this->tmpFilters = array_merge($this->tmpFilters, $filters);
```
MartinMystikJonas, line 43: This would leave filters added for single presenter/control for all following no? This kind of state can have unexpected side effects.

MartinMystikJonas: I think aybe filter variables should not be mixed with normal variables like this with global state but covered separately? Maybe AddVarTypesNode visitor can get list of variables and list of filters and do this logic there? Or maybe separate AddFilterVariablesVisitor?

MartinMystikJonas:
I think it is ok but I thiunk filter variables logic should be moved from VAriableCollectors somewhere in Compiler node visitors to avoid global state. And when I think about it makes more sense to keep this logic close together inside compiler.

src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
lulco: Unfortunatelly it isn't. I can’t find one place where to add it

MartinMystikJonas: It should be in getFilters of Latte2Compiler I think. From there all other services should get it

src/VariableCollector/DynamicFilterVariables.php
lulco: I know this is bad practice. I'm reseting it at the end of collect method. It's temporary. There will be no global variables / filters / components / forms because of standalone templates (like new Engine() for newsletters). We will collect all types in resolvers. I have to create issue for that.

lulco:
> I think it is ok but I thiunk filter variables logic should be moved from VAriableCollectors somewhere in Compiler node visitors to avoid global state. And when I think about it makes more sense to keep this logic close together inside compiler.

It has to be somewhere where we will know what kind of template we will render - if it is DefaulTemplate, we will use actual data. But for another templates there will be other variables etc. I have to analyse it a little, because I don’t use it in my projects

MartinMystikJonas:
@lulco I mean only fictinal variables for filters. They are not exactly variables but just placeholders to allow analysis. And now it showed mixing them with variables is complicated. I would just scratch DynamicFilterVariables from VariableCollectors and moved this logic (basically 1:1) to new AddDynamicFilterVariablesNodeVisitor that would behave similar to AddVarTypesNodeVisitor but will add filter palceholder variables.

This way placeholder avriables would not mix with regular variables and all these filter analysis logic would be encapsulated together in compiler.

Other VariableCollector related logic we can rework later.

PR #162, opened 2022-12-19
LatteContextAnalyser and collectors rework

### LINKED ISSUE #158: Replace FileAnalyser calls with NodeScopeAnaylser in data collecting

MartinMystikJonas:
Resolves #158 

- [x] Can we choose better name for LateContextData (new alternative of CollectedData for our own collectors)
- [x] I think we should separate classical PHPStan collectors (currently only ResolvedNodeCollector) and our LatteContextCollectors to separete namespaces. What about Colelctor and LatteContext? (including related Finders and ValueObjects)
- [x] We can remove serialization logic from latte context ValueObjects and simplify them - but it might be useful in future for caching or parallel processing. I do not know what do you think about keeping unused code just in case. I would remove it and recover from git when needed.


src/Analyser/LatteContextAnalyser.php
```diff
@@ -0,0 +1,152 @@
...
+    private NodeScopeResolver $nodeScopeResolver;
+
+    private Parser $parser;
+
+    private ReflectionProvider $reflectionProvider;
+
+    private FileHelper $fileHelper;
+
+    /** @var AbstractLatteContextCollector[] */
+    private $collectors;
```
lulco: add array typehint

src/Analyser/LatteContextAnalyser.php
```diff
@@ -0,0 +1,152 @@
...
+                }
+                $collectedRelatedFiles = array_merge($collectedRelatedFiles, $fileResult->getCollectedData(CollectedRelatedFiles::class));
+            }
+
+            $processedFiles = [];
+            $relatedFiles = [];
+            foreach ($collectedRelatedFiles as $collectedRelatedFile) {
+                $processedFiles[] = $collectedRelatedFile->getProcessedFile();
+                $relatedFiles[] = array_filter($collectedRelatedFile->getRelatedFiles(), function (string $file) {
+                    return pathinfo($file, PATHINFO_EXTENSION) === 'php'; // TODO brat z configu phpstanu
```
lulco: this was an old protection against *.stub files, now they should not appear, we can remove this array_filter

src/Analyser/LatteContextData.php
```diff
@@ -0,0 +1,66 @@
...
+
+use Efabrica\PHPStanLatte\Collector\ValueObject\CollectedValueObject;
+use PHPStan\Analyser\Error;
+
+class LatteContextData
+{
+    /**
+     * @var array<Error>
+     */
+    private $errors;
```
lulco: all properties should have typehint (if possible)

src/Collector/FilterCollector.php
```diff
@@ -31,28 +28,25 @@ final class FilterCollector extends AbstractCollector
...
+        parent::__construct($nameResolver, $reflectionProvider);
         $this->typeResolver = $typeResolver;
         $this->templateTypeResolver = $templateTypeResolver;
         $this->valueResolver = $valueResolver;
     }
 
     public function getNodeType(): string
     {
-        return Node::class;
+        return MethodCall::class;
```
lulco, line 44: If we want to open collectors, all of them should be Node::class.

But if MethodCall is here, it should be also on collectData @param and we can remove condition `if (!$node instanceof MethodCall) {` because it is always true

src/Rule/LatteTemplatesRule.php
```diff
@@ -181,7 +179,7 @@ private function analyseTemplates(array $templates, Scope $scope, array &$errors
                         $includedTemplatePath,
                         $actualClass,
                         $actualAction,
-                        array_merge($collectedTemplateRender->getVariables(), $template->getVariables()),
+                        VariablesHelper::merge($template->getVariables(), $collectedTemplateRender->getVariables()),
```
lulco, line 177: in fact, in included files there is `+` not `array_merge` so we should have variable names as keys and use the same opoerator here (I think I saw it somewhere you add the key for variables array)

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -498,17 +498,17 @@ public function testRecursion(): void
                 'recursion.latte',
             ],
             [
-                'Dumped type: 9|10',
```
lulco, line 501: how the type has changed?

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -590,7 +590,7 @@ public function testTrait(): void
                 'trait.latte',
             ],
             [
-                'Dumped type: mixed',   // TODO figure out how to change scope for assign in method
+                'Dumped type: array{\'foo\', \'bar\', \'baz\'}',
```
lulco, line 593: 👍

lulco:
I like it. Also like we get rid of some static calls and some other stuff.


> * [ ]  Can we choose better name for LateContextData (new alternative of CollectedData for our own collectors)

LatteContextData is OK for me now

> * [ ]  I think we should separate classical PHPStan collectors (currently only ResolvedNodeCollector) and our LatteContextCollectors to separete namespaces. What about Colelctor and LatteContext? (including related Finders and ValueObjects)

Separate to other namespace sounds reasonable

> * [ ]  We can remove serialization logic from latte context ValueObjects and simplify them - but it might be useful in future for caching or parallel processing. I do not know what do you think about keeping unused code just in case. I would remove it and recover from git when needed.

Remove unused code please, that's why we have git, and maybe in time we will need it, we will implement it other way (maybe better),  and we will not have to maintain this code.

> * [ ]  Maybe change latte context resolvers to support multiple nodes at oncce like rector? To avoid unnecessary calls on irrelevant nodes.

How would you implement multiple nodes without calls on all nodes if Collector supports only one natively?

> * [ ]  Create LatteContextCollectorRegistry top handle logic of searching for relevant collectors for given node. Similar to PHPStan\Collectors\Registry

What is the usage?

> * [ ]  Maybe create common ancestor for value objects that have class/method context to avoid duplication

You mean like class / method where the variable / filter / form is called etc? Is it the same in all value objects? I don't think so.

src/Analyser/LatteContextData.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @var array<class-string, CollectedValueObject[]>
+     */
+    private $collectedDataByType = [];
+
+    /**
+     * @param array<Error> $errors
+     * @param array<CollectedValueObject> $collectedData
+     */
+    public function __construct(array $errors, array $collectedData)
```
lulco: Why errors is first argument, aren't collected data more important?

MartinMystikJonas:
> > * [ ]   Maybe change latte context resolvers to support multiple nodes at oncce like rector? To avoid unnecessary calls on irrelevant nodes.
> 
> How would you implement multiple nodes without calls on all nodes if Collector supports only one natively?

Now we do not use PHPStan collectors and their interface. We can define interface for LatteContextCollectors as we want. One way to implement it is LatteContextCollectorRegistry (see next point)

> > * [ ]   Create LatteContextCollectorRegistry top handle logic of searching for relevant collectors for given node. Similar to PHPStan\Collectors\Registry
> 
> What is the usage?

Instead iterating over all collectors for all nodes you create Registry that ask collectors at initialization time what nodes they support. Then for every node you ask registry what collectors support that node and interate only over these collectors. It can improve performance.

> > * [ ]   Maybe create common ancestor for value objects that have class/method context to avoid duplication
> 
> You mean like class / method where the variable / filter / form is called etc? Is it the same in all value objects? I don't think so.

Not in all of them but in most of them. So I would keep general CollectedValueObject But create something like LocatedCollectedValueObject with common logic for value objects defined for context. Main reason is that logic of finding variables/components/forms in method call chains are duplicated in finders. If we creste common abstraction it could be deduplicated. BUt maybe leve this for some later refactoring when CollectedValueObject interfaces are stable.

src/Analyser/LatteContextData.php
MartinMystikJonas: I copied it for FileAnalyserResult and did not changed it. But agreed tht in our use case collected data is more important and should be firt parameter.

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
MartinMystikJonas: It was fixed by above replacement of array_merge with VariablesHelper::merge. Original version mixed all variable in single array instead of replacing merging them and later they were unionised when outputed to template. Original behaviour was wrong (variable remembered all previous values in include-chain) but I was not sure why. Now I found it and fixed it.

src/Rule/LatteTemplatesRule.php
MartinMystikJonas: For associative arrays (with variable names as keys) operator `+` and `array_merge` behaves the same (except swaped priorities of conflicting keys). Simple array_merge gives wrong result because variables are list without names as keys. After this change it behaves just like `+` used in compiled code.

src/Collector/FilterCollector.php
MartinMystikJonas: This is small optimization that prevents calling collector for nodes that would be droped anyway. We can always go gor nore general node if collector would need to process them.

MartinMystikJonas:
> Separate to other namespace sounds reasonable

So LatteContext? I will seaprate it

MartinMystikJonas:
@lulco We might even get rid of TypeSerializer now. It is not needed for ResolvedNOdeCollector and latte context collectors do not need serialization at all. Agreed?

lulco:
> @lulco We might even get rid of TypeSerializer now. It is not needed for ResolvedNOdeCollector and latte context collectors do not need serialization at all. Agreed?

Sure. Remove all unused code

MartinMystikJonas:
@lulco Ok. Just two other questions so I can finish this. 

1) Is `LatteContext` good name for new namespace? 
2) In new `LatteContext` namespace I would preffer to have value objects directly in LatteContext (for example `\LatteContext\CollectedVariable`) nad have sub-namespaces `LatteContext\Finder`, `LatteContext\Collector`. Do you agree or do you want to keep same structure as currently in Colector?

lulco:
> @lulco Ok. Just two other questions so I can finish this.
> 
> 1. Is `LatteContext` good name for new namespace?

Is latte context the namespace for collected data or the one used in Template?
> 2. In new `LatteContext` namespace I would preffer to have value objects directly in LatteContext (for example `\LatteContext\CollectedVariable`) nad have sub-namespaces `LatteContext\Finder`, `LatteContext\Collector`. Do you agree or do you want to keep same structure as currently in Colector?

Different is only no dir for ValueObjects right?

MartinMystikJonas:
`LatteContext` for collected data `Template` for data in template. Any ideas for better names? 

No dir instead of `ValueObject` for value objects and `Collector` instead of no dir for collectors

lulco:
> `LatteContext` for collected data `Template` for data in template. Any ideas for better names?

I was thinking about this longer. What is the difference? In forms there was basically no need to add Template data and collected data - they works only with collected one. I see the difference only in class / method additional data. At the beginning there was no Collected data, only those in Template dir, but this package is evolving :) What about to use only "TemplateContext” in both sides - when they are collected and also when they are used?

> No dir instead of `ValueObject` for value objects and `Collector` instead of no dir for collectors

I would like to have ValueObjects in some dir just don’t like the name value object 🤷 any idea? Without dir would stay only classes which are not “categorizable”.

MartinMystikJonas:
> > `LatteContext` for collected data `Template` for data in template. Any ideas for better names?
> 
> I was thinking about this longer. What is the difference? In forms there was basically no need to add Template data and collected data - they works only with collected one. I see the difference only in class / method additional data. At the beginning there was no Collected data, only those in Template dir, but this package is evolving :) What about to use only "TemplateContext” in both sides - when they are collected and also when they are used?

I see a difference between them

Template data - represents data that are used to properly compile and analyse template
Collected data - are intermediate objects used to collect partial information from which tempalte data are constructed by resolvers and finders. It often represents just partial concepts of template data (like TemplatePath), concepts totally unknown to templates (MethodCall or RelatedFile) or contains same data but with added context (like Variable with information about class/method where it was collected and if it was assignment of explicitly declared type, or mentioned Form)

Basically template data are end result of resolvers while collected data are their input. Maybe naming should somehow represent this? 

What about TemplateContext for tempalte and CollectedLatteData, LatteData or LatteInfo for collected data?

I would preffer them to be separate to avoid mixing tempalte related data with lower level concepts. But collected data should be created as wrappers of tempalte data when possible - CollectedForm will have Tempalte\Form property with added method/class info. FormField can be only in Template. 
 
> > No dir instead of `ValueObject` for value objects and `Collector` instead of no dir for collectors
> 
> I would like to have ValueObjects in some dir just don’t like the name value object 🤷 any idea? Without dir would stay only classes which are not “categorizable”.

What about simply Item, Entity or Data?

lulco:
> > > `LatteContext` for collected data `Template` for data in template. Any ideas for better names?
> > 
> > 
> > I was thinking about this longer. What is the difference? In forms there was basically no need to add Template data and collected data - they works only with collected one. I see the difference only in class / method additional data. At the beginning there was no Collected data, only those in Template dir, but this package is evolving :) What about to use only "TemplateContext” in both sides - when they are collected and also when they are used?
> 
> I see a difference between them
> 
> Template data - represents data that are used to properly compile and analyse template Collected data - are intermediate objects used to collect partial information from which tempalte data are constructed by resolvers and finders. It often represents just partial concepts of template data (like TemplatePath), concepts totally unknown to templates (MethodCall or RelatedFile) or contains same data but with added context (like Variable with information about class/method where it was collected and if it was assignment of explicitly declared type, or mentioned Form)
> 
> Basically template data are end result of resolvers while collected data are their input. Maybe naming should somehow represent this?
> 
> What about TemplateContext for tempalte and CollectedLatteData, LatteData or LatteInfo for collected data?
> 
> I would preffer them to be separate to avoid mixing tempalte related data with lower level concepts. But collected data should be created as wrappers of tempalte data when possible - CollectedForm will have Tempalte\Form property with added method/class info. FormField can be only in Template.
> 

I see the difference, just asking if it isn't too complicated structure for such simple extension :) maintain 2x more classes it needs (my actual feeling). OK let’s keep both.

> > > No dir instead of `ValueObject` for value objects and `Collector` instead of no dir for collectors
> > 
> > 
> > I would like to have ValueObjects in some dir just don’t like the name value object 🤷 any idea? Without dir would stay only classes which are not “categorizable”.
> 
> What about simply Item, Entity or Data?

CollectedData?

MartinMystikJonas:
So

- `Collector`
  - `CollectedData`
  - `Collector`
  - `Finder`

- `LatteContext`
  - `CollectedData`
  - `Collector`
  - `Finder`

- `Template`

OK?

lulco:
Ok

MartinMystikJonas:
Should be ready for merge

lulco (approved):

Issue #119, opened 2022-12-06
Differentiate between sometimes and always called methods


MartinMystikJonas:

MartinMystikJonas:
Partially solved in #167 but we should look for better way if possible

PR #167, opened 2022-12-23
NetteApplicationUIPresenterCollector to collect presenter specific renders and method calls

### LINKED ISSUE #91: Presenters: Different template and render method when action uses setView
Presenter action can call setView to use different render method and template than defaukt derived from action name.

Probably add ViewChangeCollector and evaluate it in presenter resolver.

### LINKED ISSUE #102: Resolve renders from calls to sendTemplate() in presenters

MartinMystikJonas:
Resolves #91 
Resolves #102 
Basic implementation of #119 

Summary:
- `CollectedMethodCall` have information if call is performed always (is directly in method body) or sometimes (is nested) - not perfect detection but for basic use cases it is enough - I will try to finde better mechanism later
- `CollectedMethodCall` can optionally store parameters (to be used by resolvers)
- `MethodCallFinder` now returns array of `CollectedMethodCall` instead of just class/method names of called methods to allow use of new information collected for method calls (params, conditionaly/always called)
- New `NetteApplicationUIPresenterCollector` to collect `Presenter` specific renders and method calls for use by `NetteApplicationUIPresenter` resolver
- `sendTemplate` call in `Presenter` is considered as render call
- `NetteApplicationUIPresenter` resolver change of view by calling `setView` in action method

Limitations:
- if setView is called inside condition default view matching action name is also rendered even in cases when setView is called in every brach of condition - I found no way to detect this - may left some false positives

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 - Annotation `@phpstan-latte-ignore` can be used to ignore render calls, variable assignments, component creation, whole methods or classes.
 - Annotation `@phpstan-latte-template` can be used to specify what template is used to render.
 - Annotation `@phpstan-latte-var` can be used to specify what variables are available in tempalte.
-
+- Resolves calls to `setView` and `sendTemplate` in presenters
+ 
 ## [0.2.0] - 2022-12-06
 ### Changed
 - Used collectors to find all variables, components, method calls and templates to analyse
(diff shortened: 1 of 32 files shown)


MartinMystikJonas:
TODO
- [ ] `MethodCallFinder` should have option to return all called methods recursivelly with added logic that non-conditional call in conditionally called method is also considered conditionally called
- [ ] `setView` calls should be detected also in methods called from action method (requires previous)

lulco (approved):
I don't see use case when setView is used in `startup()` method, is it covered?

MartinMystikJonas:
It is not. Never seen such use case. But it is possible I think. I will check it.

MartinMystikJonas:
Added ability to resolve setView in startup and merged.

PR #170, opened 2022-12-25
Fixed CalledClassResolver to properly handle generics

MartinMystikJonas:


lulco (approved):
Good to know there is such method :)

lulco:
Can we add some regression test?

Issue #142, opened 2022-12-11
Custom form field resolving


MartinMystikJonas:
Possibility to use FormFinder to collect form fields from given method. So resolvers can explicitly add additional forms or additional form fields to autoamticaly collected forms.

@MartinMystikJonas TODO: Prepare test case

MartinMystikJonas:
This is one of two last things I need to solve to have no errors in our real world projects.

I was thinking about it and I think we need to make these changes:

1) Form controls (`CollectedFormField`) creation should be collected everywhere (similar to component creation) based only on called type is Form. Now it is only collected in `createComponent` that returns `Form`
2) FormFinder should be replaced by `FormFieldFinder` able to do recursive search for added form controls `CollectedFormField` (even in methods called from `createComponent`)
3) Resolving available forms should be moved to Resolvers - resolver should identify forms (based on existence of `createComponent` returning Form, but can use different methods if needed), and then create `CollectedForm` with given `CollectedFormFields` (that can be collected from many places)

MartinMystikJonas:
Point is that we need to be able to do something like this in Resolvers: 

```php
  $fields = array_merge(
      $this->formFieldFinder->find($class, 'createComponentForm')
      $this->formFieldFinder->find($class, 'initForm')
  );
$forms = new Form($name, $type, $fields);
```

MartinMystikJonas:
I am not sure where logic to detect default existing forms for given control/presenter should be. I think it should be in Resolvers so resolvers have full control about it. But at the same time I feel it should be independent.

Maybe have `FormCollector` and  `FormFinder` that would return only places where forms are created (basically only `createComponent` with return type `Form`) and will not attempt to return form fields (current implementation of analysist stmts of method have same problem with frozen scope we discussed earlier). This will be `CollectedForm` - just name and type without fields.

And separate 'FormFieldCollector' and 'FormFierldFinder' that collects `CollectedFormFields` (but without information about what Form it belongs to).

ANd then Resolver will create `Form` including `FormField`s from information from collected context data from finders.

MartinMystikJonas:
So to summarise it:

`CollectedForm` - class, method, name, type
`CollectedFormField` - class, method, name, type

`Form` - name, type, form fields
`FormField` - name, type

`FormCollector` + `FormFinder` - collects forms created in class/method (without fields)
`FormFieldCollector` + `FormFieldFinder` - collected form fileds added to form in given method (without knowing to what form it is added)

[...]

PR #171, opened 2022-12-26
Test case for header variables undefined error

MartinMystikJonas:
When variable is defined in tempalte header code for variable assignment is placed inside `prepare` method of template. Other methods in template therefore have no idea variable exists during analysis. Pepare method is called before calling odf template to modify content od parameters passed to template.

Not sure how to best solve this. Maybe copy content of prepare method on start of every template method just after parameters var annotations?

### DIFF
--- a/tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Variables/other.latte
+++ b/tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Variables/other.latte
@@ -1,7 +1,10 @@
+{var $headerVar = "something"}
+
 {block content}
 
 {$fromOtherAction}
 {$fromRenderDefault}
 {$nonExistingVariable}
+{$headerVar}
 
 {include @partial.latte, renamedFromOtherAction: $fromOtherAction}
--- a/tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
+++ b/tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
@@ -95,12 +95,12 @@ final class LatteTemplatesRuleForPresenterTest extends LatteTemplatesRuleTest
             ],
             [
                 'Variable $fromRenderDefault might not be defined.',
-                4,
+                6,
                 'other.latte',
             ],
             [
                 'Variable $nonExistingVariable might not be defined.',
-                5,
+                7,
                 'other.latte',
             ],
             [
--- a/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Variables/SomeControl.php
+++ b/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Variables/SomeControl.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables;
+
+use Nette\Application\UI\Control;
+
+final class SomeControl extends Control
+{
+    public function render(): void
+    {
+        $this->template->a = 'a';
+        $this->template->render(__DIR__ . '/default.latte');
+    }
+}
--- a/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Variables/default.latte
+++ b/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Variables/default.latte
@@ -1,0 +1,4 @@
+{var $headerVar = 'something'}
+
+{$nonExistingVariable}
+{$headerVar}
--- a/tests/Rule/LatteTemplatesRule/SimpleControl/LatteTemplatesRuleForSimpleControlTest.php
+++ b/tests/Rule/LatteTemplatesRule/SimpleControl/LatteTemplatesRuleForSimpleControlTest.php
@@ -277,4 +277,15 @@ final class LatteTemplatesRuleForSimpleControlTest extends LatteTemplatesRuleTes
             ],
         ]);
     }
+
+    public function testVariables(): void
+    {
+        $this->analyse([__DIR__ . '/Fixtures/Variables/SomeControl.php'], [
+            [
+                'Variable $nonExistingVariable might not be defined.',
+                3,
+                'default.latte',
+            ],
+        ]);
+    }
 }


MartinMystikJonas:
```php
<?php

use Latte\Runtime as LR;
/** 
* Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl::render
* @property PHPStanLatteTemplate_a5b95bc7141055e7cd566adf4a92665d_global $global
*/
final class PHPStanLatteTemplate_a5b95bc7141055e7cd566adf4a92665d extends Latte\Runtime\Template
{
    public function main(array $ʟ_args) : void
    {
        extract($this->params);
        /** @var Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl $presenter */
        /** @var Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl $control */
        /** @var 'a' $a */
        /** @var string $baseUrl */
        /** @var string $basePath */
        /** @var Latte\Runtime\FilterInfo $ʟ_fi */
        /** @var array<string> $ʟ_tag */
        /** @var array $ʟ_if */
        /** @var array $ʟ_ifc */
        /** @var array $ʟ_try */
        /** @var array $ʟ_loc */
        /** @var Nette\Security\User $user */
        /** @var Nette\Application\UI\Presenter $presenter */
        /** @var Nette\Application\UI\Control $control */
        /** @var array<stdClass> $flashes */
        /** @var Nette\Localization\Translator $__filter__translate */
        unset($ʟ_args);
        if ($this->global->snippetDriver?->renderSnippets($this->blocks[self::LayerSnippet], $this->params)) {
            return;
        }
        echo "\n";
        /* line 3 */
        echo LR\Filters::escapeHtmlText($nonExistingVariable);
        echo "\n";
        /* line 4 */
        echo LR\Filters::escapeHtmlText($headerVar);
        echo "\n";
    }
    public function prepare() : array
    {
        extract($this->params);
        /** @var Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl $presenter */
        /** @var Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl $control */
        /** @var 'a' $a */
        /** @var string $baseUrl */
        /** @var string $basePath */
        /** @var Latte\Runtime\FilterInfo $ʟ_fi */
        /** @var array<string> $ʟ_tag */
        /** @var array $ʟ_if */
        /** @var array $ʟ_ifc */
        /** @var array $ʟ_try */
        /** @var array $ʟ_loc */
        /** @var Nette\Security\User $user */
        /** @var Nette\Application\UI\Presenter $presenter */
        /** @var Nette\Application\UI\Control $control */
        /** @var array<stdClass> $flashes */
        /** @var Nette\Localization\Translator $__filter__translate */
        /* line 1 */
        $headerVar = 'something';
        return get_defined_vars();
    }
}
/**
* @property array $coreParentFinder
* @property Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl $uiControl
* @property Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Variables\SomeControl $uiPresenter
* @property Nette\Bridges\ApplicationLatte\SnippetDriver|null $snippetDriver
* @property string|null $uiNonce
* @property array<int, Nette\Forms\Container> $formsStack
*/
class PHPStanLatteTemplate_a5b95bc7141055e7cd566adf4a92665d_global extends \stdClass
{
}
```

MartinMystikJonas:
Prepare is called here https://github.com/nette/latte/blob/61d72cfabd6764880942c79d3cd049a77d056d95/src/Latte/Runtime/Template.php#L139

lulco:
Hm, this should be already solved by compiler itself or not? If compiled template hasn't it in all methods then static analysis is right if report missing variable. Or am I missing something? Is it working in real world app? How?

MartinMystikJonas:
Thete is a bit magic in it. `prepare` method is called before template is rendered. Variables created in `prepare` method are collected by `get_defined_vars` and then added to `$this->params`. So they are present after call to `extract`.

lulco:
I see, then probably copy body of prepare (without return) to all other methods is the best solution.

MartinMystikJonas:
Could you try to write such nodevisitor? I am not sure how to do it correctly.

lulco:
Yes, I'll try. About correctness - I'm not sure if I wrote them correct, they just work :)

MartinMystikJonas:
Best kind of correct 😉

lulco:
"Great", in latte 2.x `prepare()` method doesn't return anything ;) in that version `main()` returns data, see https://github.com/nette/latte/blob/v2.11/src/Latte/Runtime/Template.php#L201

MartinMystikJonas:
But it seems this is used in Latte 2 only to pass variables to parent (extended) template. In Latte 3 prepare method is used always.

MartinMystikJonas:
However we should sokve evaluation of extended temates somehow. But I think it woukd be better handled by same mechanism as includes. We would just have to pass all variables defined at the end of main method to extended template. But it can be easily obtained from Scope in ExecutionEndNode of main method of compiled template.

MartinMystikJonas:
When this and #142 is solved then I would have only few dozens errors in our biggest real world project that has to be solved and most of them are real bugs or missing checks for invalid input (that does not happen but could)

It seems we are are really close to fully working solution for real projects :-)

lulco:
I've analysed this and I got to this:
This is SOOO edge case for me (I've never use it), but I have to add 3 NodeVisitors and change next 2-3 existing NodeVisitors to make it work. It will get too much time I don't have now. Maybe little bit later.

MartinMystikJonas:
Damn I hoped it would not be that complicated. I use it only in few places (basically just to define constant used on multiple places to make modifications easier) so I will try to find some workaround for now.

MartinMystikJonas:
Do you have any solution draft in progress? Maybe I can try to finish it.

lulco:
Main problem is that latte 2 and latte 3 use it differently, in latte 3 it is just types and `return get_definded_vars()`, but in latte 2 there are some ifs and method calls etc. and also `return get_definded_vars()`, so I can't easily find what code to copy. So I have to add some marker to these methods (first visitor), then copy code from method start to marker (second visitor), then remove mark and figure out how to keep comments - comment is part of some Expression, and if only comments are in prepare method with `return get_definded_vars()` I can't remove that return (third visitor). Maybe marker should be some harmless expression like `$___marker___ = $this;` which does nothing but allows us to keep comments above it (and also I will have only two visitors because I will not do cleanup).

Let me few days (or 2 weeks) to handle it. I am still on vacation so I'm not working on it now.

MartinMystikJonas:
Ok I will use workaround for now and we can solve it later.

MartinMystikJonas:
Solved by #183 and #184

PR #175, opened 2022-12-29
Annotation @phpstan-latte-component

### LINKED ISSUE #113: RFC: Annotations
There are cases that cannot be resolved autimatically by static analysis. In these casses new annotations cound be used to guide resolvers to analyse latte correctly.

### `@phpstan-latte-ignore`
Allowed context: variable assign, method call, render call, method, class

```php
class MyControl extends Control
{
    public function render(string $param): void {
        if ($param) {   
            /** @phpstan-latte-ignore */
            $this->template->something = $param; // <-- this variable will not be collected (by default all template variables assigned inside method are collected)
        } else {
            $this->render(__DIR__ . '/MyControl.latte');
        }
    }
}
```

```php
class MyControl extends Control
{
    public function render(string $param): void {
        if ($param) {       
            /** @phpstan-latte-ignore */   
            $this->render(__DIR__ . $param); // <-- this will not be attepted to resolve (reported as expression that cannot be resolved)
        } else {
            $this->render(__DIR__ . '/MyControl.latte');
        }
    }
}
```

```php
class MyControl extends Control
{
    /** @phpstan-latte-ignore */ // <-- whole method is ignored by resolvers and collectors
    public function render(string $param): void {
        $this->template->something = $param; // <-- this variable will not be collected
        if ($param) {       
            $this->render(__DIR__ . $param); // <-- this will not be resolved
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- this will not be resolved
        }
    }
}
```

```php
class MyControl extends Control
{
    public function render(string $param): void {
        /** @phpstan-latte-ignore */         
        $this->doRender(); // <-- this method call wil not be follwed by resolvers (resolving renders or collecting variables inside called method)
    }

    private function doRender(): void {
        $this->template->something = $param; // <-- this variable will not be collected when resolving render()
        if ($param) {       
            $this->render(__DIR__ . $param); // <-- this will not be resolved when resolving render()
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- this will not be resolved when resolving render()
        }
    }
}
```

```php
/** @phpstan-latte-ignore */ // <-- whole class is ignored by resolvers and collectors
class MyControl extends Control
{
    public function render(string $param): void {
        $this->template->something = $param; // <-- this variable will not be collected
        if ($param) {       
            $this->render(__DIR__ . $param); // <-- this will not be resolved
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- this will not be resolved
        }
    }

    public function renderAlternative(string $param): void {
        $this->template->something = $param; // <-- this variable will not be collected
        if ($param) {       
            $this->render(__DIR__ . $param); // <-- this will not be resolved
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- this will not be resolved
        }
    }
}
```

### `@phpstan-latte-template`
Allowed context: render call, method, class

Value specifies what template or templated should be resolved and checkid in given context.

Multiple annoations can be used on single element or at different levels. When used all of them will be resolved.

Can use wildcard `*` to match multiple files
Can use placeholders:
`{dir}` = `__DIR__` (example `/app/dir/controls`)
`{file}` = `__FILE__` (example `/app/dir/controls/MyControl.php`)
`{fileName}` = `basename(__FILE__)` (example `MyControl.php`)
`{fileNameNoSuffix}` = file anme with suffix removed (example `MyControl`)
`{classShortName}` = class short name withnout namespace (example `MyControl`)

Question: regex support?

```php
class MyControl extends Control
{
    public function render(string $param): void {
        /** @phpstan-latte-template {dir}/templates/{classShortName}.*.latte */
        $this->render($this->getTemplatePath($param)); // <-- will resolve all templates matching given pattern and will not report error of expression that cannot be evaluated
    }
}
```

```php
class MyControl extends Control
{
    /** @phpstan-latte-template {dir}/templates/{classShortName}.*.latte */ // <-- will resolve all templates matching given pattern + anything resolved from method body
    public function render(string $param = null): void {
        if($param !== null) {     
            $this->render($this->getTemplatePath($param)); // <-- will be resolved normally (will report error of unresolved expression)
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- will be resolved normally
        }
}
```

```php
class MyControl extends Control
{
    /** @phpstan-latte-template {dir}/templates/{classShortName}.*.latte */ // <-- will resolve all templates matching given pattern + anything resolved from method body
    public function render(string $param = null): void {
        if($param !== null) {     
            /** @phpstan-latte-ignore */
            $this->render($this->getTemplatePath($param)); // <-- will be ignored so no error now
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- will be resolved normally
        }
    }
}
```

```php
/** @phpstan-latte-template {dir}/templates/{classShortName}.*.latte */ // <-- will resolve all templates matching given pattern in all class render methods + anything resolved from method bodies
class MyControl extends Control
{
    public function render(string $param = null): void {
        if($param !== null) {     
            /** @phpstan-latte-ignore */
            $this->render($this->getTemplatePath($param)); // <-- will be ignored so no error now
        } else {
            $this->render(__DIR__ . '/MyControl.latte'); // <-- will be resolved normally
        }
        // < -- all tempaltes matching {dir}/templates/{classShortName}.*.latte will be resolved from context of this method
    }
    public function renderAlternative(string $param = null): void {
        // < -- all tempaltes matching {dir}/templates/{classShortName}.*.latte will be resolved from context of this method too
    }
}
```

### `@phpstan-latte-var`
Allowed context: render call, method, class

Defines variable type passed to template in given context.

Multiple annoations can be used on single element or at different levels. 

When same variable name is defined at multiple levels then lowest level definition is used. So if class annotationd defines variable as `string|int` but annotation on render call specifies it as `string` then in template it will be evaluated as `string`.

If type `void` is used then tempalte will be undefined in template.

Annotations takes precedence over anything collected from code. Even when used on different level. So when class annoation defines variable type it will overwrite variable type collected in method calls or render call parameters.

Question: Is this behaviour expected? Or should annotation overwrite only same level?

TODO examples

### `@phpstan-latte-component`

Same behaviour as `@phpstan-latte-var` but for components.

MartinMystikJonas:
Resolves #113


src/Collector/Collector/ResolvedNodeCollector.php
```diff
@@ -42,16 +42,16 @@ public function getNodeType(): string
...
-            return null;
-        }
         $resolvedNodes = [];
         foreach ($this->latteTemplateResolvers as $latteTemplateResolver) {
             $resolvedNode = $latteTemplateResolver->collect($node, $scope);
             if ($resolvedNode !== null) {
                 $resolvedNodes[] = $resolvedNode;
             }
         }
+        if (count($resolvedNodes) > 0 && $this->lattePhpDocResolver->resolveForNode($node, $scope)->isIgnored()) {
```
MartinMystikJonas, line 52: This was moved because I found out that calling it for every node (even when no resolver matches it) can have  memory and performace impact

MartinMystikJonas:
Once again a bigger PR but most of it is tests and I hope this is last one that big :-)

lulco (approved):

PR #176, opened 2022-12-29
Fixed detection of standalone presenter templates

MartinMystikJonas:


src/LatteTemplateResolver/NetteApplicationUIPresenterStandalone.php
```diff
@@ -24,7 +24,7 @@ protected function getTemplatePathPatterns(ReflectionClass $reflectionClass, str
 
         return [
              $dir . '/templates/' . $presenterName . '/([a-zA-Z0-9_]*?).latte',
-             $dir . '/templates/' . $presenterName . '\.([a-zA-Z0-9_]?).latte',
+             $dir . '/templates/' . $presenterName . '\.([a-zA-Z0-9_]*?).latte',
```
lulco: Shouldn’t be `+` there? And also few lines above too?

MartinMystikJonas: Yeah that would be better

lulco (approved):

PR #177, opened 2022-12-29
Ignore reporting of unknown @phpstan-latte-* PHPDoc tags

MartinMystikJonas:


lulco (approved):

lulco:
This should be in php file ErrorBuilder because errors from config are removable (with ! in config)

MartinMystikJonas:
Thisnis repoeted directly by phpstan as error in php file it does not pass our error builder. Or is there way to influence phostan own errors?

lulco:
So if I have reportUnmatchedErrors turned on this will report issues? It's not good.

MartinMystikJonas:
True we should explicitly turn off reportUnmatched for these errors

lulco:
Just thinking - isn't there some “register” of available phpdoc in phpstan core? It would be great to just register new types to it.

MartinMystikJonas:
It is hardcoded constant https://github.com/phpstan/phpstan-src/blob/73bb9eb0e1f3058f8aba5ee9fb88cad58734c44f/src/Rules/PhpDoc/InvalidPHPStanDocTagRule.php#L22

lulco:
Hmm this is not so unrealistic to change to parameter 🤷

MartinMystikJonas:
Well we can draft PR for it but I am not sure it will be accepted. It is porblem only for annotations starting with `@phpstan`

lulco:
It's phpstan- so we can:
- change our anotation to `phpstanlatte-` (easy win)
- create PR to phpstan
- set report unmatch errors to false for these errors (easy win)

MartinMystikJonas:
I think option 3 is easiest and without BC break so I would go for it.

lulco:
Ok then go for it

PR #178, opened 2022-12-30
Fixed detection of standalone presenter templates when methods are ignored

MartinMystikJonas:
When presenter method is ignored resolving of template should not fall back to standalone template


lulco (approved):
Can be some test case added?

MartinMystikJonas:
Test added

PR #179, opened 2022-12-31
Fixed method call collecting outside classes

MartinMystikJonas:
Original code expected node to always have grandaprent node. It is not always true.


lulco (approved):
To be honest, I don’t know this code so I will trust you :)

PR #180, opened 2022-12-31
Improved form fields collecting

MartinMystikJonas:
See #142


lulco:
So I think we just should remove condition to class method createComponent* and collect all forms. Then use method calls in finder like in variables and components is used. That should be enough, right?

MartinMystikJonas:
@lulco Not really because method can just add form fields not create entire form. So we need to write collectors/finders to work directly with FormFields. Then we can collecto forms and populate them with relevant collected fields.  Also current implementation would fail when form field is for example in variable  - because content of createCopmponent is evaluated with frozen scope from start of the method. I described proposed solution in #142.

MartinMystikJonas:
We need to treat form fields as first class citizens same as variables  with own collector and finder. And Form colector/finder will just seatch for form creation and use formfield finder to look up fields present in that form.

lulco:
OK, you're right. Forms and fields should be collected separatelly. And what will we do with containers and groups? They will also be collected separatelly? In my project, where I test this extension, containers (and their fields) are actually last problem which reports an error, so it's critical for me to solve also this.

MartinMystikJonas:
Containers would be a challenge. Because we would need some way to remember container name while it is passed around in variable. And also when it is passed to method calls so fields addded in called method knows in which container they are. I am not sure how we can even do that in PHPStan. It would probably require some advanced magic with scope if it is even possible.

MartinMystikJonas:
For now I would aim to solve just changes in collectors nad add support for containers later. It should be possible (discussed it with ondrej on slack) but probably quite a challenge to do it properly :-) It will require custom types and TypeSpecifiyingExtensions

lulco:
Maybe type extension is correct, I'm not sure. I was thinking about "simple" collectors for fields, containers and forms. We should remember class/method where these items are added and then just find relevant fields / containers in finder and create final Form. Same as for Variables and Components, just little bit more complicated structure.

MartinMystikJonas:
Problem with containers is that we need to know what fields are in what container. And that is not easy to do with basic phpstan types. Because all containers are just Container and we do not know which one and how named. So we woukd know there are some containers and that there are some fields in containers but we woukd do not know how to pair them together

MartinMystikJonas:
For forms and fields it should be similar to variables/components. Because usually there is no reason to have multiple forms mixed in one method. But containers are different because usually you work with multiple containers in single method.

lulco:
well, yes, but you know to which container fields are added or not? I thought the problem is only with fields / containers added in external methods

MartinMystikJonas:
No you do not that is the problem. You know container nate at node where it is created but that information is immediately forgotten. When you call methods on container you know only type but not name of that container. We need a way to remember what name container have. And when container is first created and then added to form we neet to remember what fields was edded to it so we can evaluate when container is added to form.

MartinMystikJonas:
@lulco I implemented form collectors rework that supports collecting form fields across methods and adding fields by `addComponent`

src/Template/Form.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Template;
```
lulco: I would create Form namespace for Form classes / interfaces. We will need some Form, Field, Container, Group etc. also Field and Container are “Controls” (can be added to Form) then Form, Container and Group are “ControlHolders” (contains Controls). So Forms have complex structure and it should not be in Template namespace directly

lulco (commented):
I’ll check it on desktop later

src/Template/Form.php
MartinMystikJonas: Would you reffer to keep Form in Tempalte os tol level element together with other top elements or move it too to Template\Form\Form?

lulco: I think Template\Form\Form would be better to keep all form items together

MartinMystikJonas: ok

MartinMystikJonas: Same for LatteContext/CollectedData I expect

lulco: Yes

lulco (approved):

lulco (approved):

PR #181, opened 2022-12-31
Fixed evaluation of overwritten method calls

MartinMystikJonas:
When I wrote test for forms I noticed there is one thing our method call analysis did not take into account.

When child class calls parent method, and that parent method calls method on `$this` that is overwritten in child class we did not analyse child version of method but overwritten parent version of method.

Problem is that we evalute all calls directly when we encounter that call in parent method. But we should not evaluate calls to `$this` in collector. It should be evaluated in finder based on currently evaluated class - that could be child.

Similar mechanism should work for `static::method()` calls


MartinMystikJonas:
I will fix it. I drafted change and except one last error it seems to work.

MartinMystikJonas:
1) CollectedMethodCall can store called class as `this` or `static`
2) MethodCallFinder evaluates `this` or `static` as current class
3) Removed all `findByMethod` methods - because `ReflectionMethod` knows only about declaring class but not about current class - must be replaced by class+method

lulco (approved):

PR #182, opened 2023-01-02
Form node visitors

lulco:
Rewritten replace / preg_replace to node visitor


src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -0,0 +1,181 @@
...
+                return null;
+            }
+            $formNameString = $dimFetch->dim;
+            $formName = $formNameString->value;
+
+            $formClassName = $this->formClassNames[$formName] ?? null;
+            if ($formClassName === null) {
+                return null;
+            }
+            $this->actualForm = $this->forms[$formName] ?? null;
```
MartinMystikJonas, line 77: I think we should add __compilerError with `Form with name *** probably does not exists` here when form is not found

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    /**
+     * @param Node[] $nodes
+     * @return Node[]
+     */
+    public function afterTraverse(array $nodes): array
+    {
+        $componentType = '\Nette\ComponentModel\IComponent';
```
MartinMystikJonas: Would not it be better to expect Form control here instead of generic component? It almost always would be control.

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -0,0 +1,181 @@
...
+                                new Arg(new Variable('name')),
+                            ]
+                        )
+                    ),
+                ])
+                ->makePublic()
+                ->setReturnType($componentType);
+            $comment = '@return ' . $componentType;
+            foreach ($form->getFormFields() as $formField) {
+                $comment = str_replace($componentType, '($name is \'' . $formField->getName() . '\' ? ' . $formField->getTypeAsString() . ' : ' . $componentType . ')', $comment);
```
MartinMystikJonas: Maybe it would be better to use some placeholder like `%%TYPE%%` here instead of `$componentType` - theoretically `$componentType` could appear as part of namespace.

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
lulco: This is just compatibility with nette itself, unfortunatelly you can add any Nette\ComponentModel\IComponent to Form / Container. But I would prefere some Form\BaseControl here, too.

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
lulco: I have to check what it is :) this part is just copy-pasted from old code

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
lulco: 👍 good idea, just have to think what to do with rest of array fetches on non-existng form 🤔

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
MartinMystikJonas: I think we can reasonably expect that when someone uses `{input}` or `{label}` he is trying to access `BaseControl` and in other cases this definition is not even used I think.

lulco: that would be great, I'll change it to BaseControl than...

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
MartinMystikJonas: Maybe just use generic `\Nette\Forms\Form` instead of defined fictional class?

lulco: I think then there will be errors like cannot call ->getControl() / ->getLabel() on mixed. Ideal scenario would be to remove whole form part and keep just error about undefined form

lulco: I added this to features in my Forms task. Will be solved later.

MartinMystikJonas (approved):

PR #183, opened 2023-01-02
Copy defined vars

lulco:


MartinMystikJonas:
I will check it in more detail tomorrow. Jsut few quick questions?

Why do you need to first add marker instead of evaluating where to stop directly in CopyDefinedVarsToOtherMethods? Is there some catch I am missing?

Would not it be easier to copy entire content of prepare/main method and just skip/replace returns?

Why visitors that add variables skip prepare/main? Content of prepare/main can rely on some passed parameters.

lulco:
I decided to go with two visitors (add marker and copy vars) because when I tried it in one visitor I had only statements of class method, and there I had to check more complicated conditions.

It's not ready, but you can check because I'm bit lost with it. I can't find some reasonable condition when to add marker / stop copying. I have no idea what can be in "head", how can be variable created there and how it is compiled.

Copy the entire prepare / main  and "just" skip return was my first try back in december, but it didn't work. In our test cases I found at least 5 - 6 different expressions which have to be skipped (not copied). Some of them because there is no reason to copy them but others (more important) because they duplicate errors from main to other methods. For example assigning non existing component:
```$_tmp = $this->global->uiControl->getComponent("foo");```
but also ```PHPStan\dumpType($foo)``` or begin of form ```echo Nette\Bridges\FormsLatte\Runtime::renderFormBegin($foo)```

Now I started to work with latte 3 and it is completely different.

Also I screwed types in block methods (where the type of argument of block overrides global variable type). Because in previous version we've added types to each method separately, now we just copy  main / prepare body (as I'm thinking now, I can copy it before variable php docs are added, but the problem with marker always stays here).

MartinMystikJonas:
OK I can try to play with it.

Problem with duplicated errors would there be always because if head for example uses undefined parmeter there is no way to avoid it. Maybe best approach would be to ignore duplicate errors.

I am thinking about alternative approach but with lower performance. We can compile template, run analyser and collect variables defined at the end of prepare/main, add it to parameters and then compile again with updated parameters and do final analysis. It would simulate exactly how it behaves in latte but requires two passes.

lulco:
What is "end of prepare/main"? That's exactly my problem here which I need to solve. There is no marker where it starts and where it ends.

MartinMystikJonas:
In Latte 3 prepareis separate method. I am not sure how it is done in Latte 2 I would need to check.

lulco:
> Why do you need to first add marker instead of evaluating where to stop directly in CopyDefinedVarsToOtherMethods? Is there some catch I am missing?

I just recall why I need that marker. I have to add some statement which will hold comments (var php docs), because if there is only 
`return get_definded_vars()` and I remove it, php docs are removed too (because they are part of node under them).

lulco:
Anyway, I reworked it little bit, tests passed, it works, we just will change the marker condition if needed. We would need some test cases. Maybe you will find some in your projects, but actual cases work, so if you are ok with it, we could merge

MartinMystikJonas:
I still do not like use of marker. If I understand correctly problem was that varriable types doc comments above marker was lost when copiing method content, because thex were attached to skipped statement. But attaching them to first statement after extract is imho mistake. That ststrement could be something like foreach where only doc comments for variables defined in foreach are allowed. I think we should move variable types above extract. That would solve both problems.

MartinMystikJonas:
I would like to play with it some more. Would you prefer to merge it as is and let me refactor it later or should I create new PR branched from this with my rafactoring?

lulco:
> I still do not like use of marker… attaching them to first statement after extract is imho mistake.

This is big missunderstanding how marker works. It is not first statement after extract.

It is added before first “header ending” statement.

lulco:
@MartinMystikJonas It seems tests are little bit slower. Probably we introduced some performance issue

MartinMystikJonas:
How much slower? Except more node visitors that has to work with tempalte I do not think there is significant reason for slow down

MartinMystikJonas:
I tested it on my PC and both before and after this commit I have 32s to run test suite. Might be just random slow down on GitHub side. Can you check it on your computer?

lulco:
> How much slower? Except more node visitors that has to work with tempalte I do not think there is significant reason for slow down

Yesterday I checked it on my PR and it was 1m37s, now it is more than 2m.

Yes there are more visitors and one more priority (75), so one more time where we have to parse nodes, but it was also yesterday. Hope it is github slowing down. Also we can try to change priority to 50 and run the copy together with blocks. I think they are not conflicted and can be run together.

MartinMystikJonas:
Looking at history https://github.com/efabrica-team/phpstan-latte/actions/workflows/phpunit.yml it seems time randomly jumps between 1:30 and 3:00. Might depends on what executor machine GitHUb uses.

MartinMystikJonas:
There is 1:30 and 2:52 for commits where only difference is removed unused use. Looks quite random here. Probably we should monitor performance on local.

PR #184, opened 2023-01-03
Copy defined vars v2

MartinMystikJonas:
* Removed need for marker
* Simplified end of header condition - copy until first return or defined vars or end of relevant block. I think we should always copy entire header part. Advanced conditions that tried to prevent duplicate errors could also skip relevant parts of header when header items are in wrong order. And some duplicate errors (like use of undefined variable in tempalte header) could not be avoided at all.
* Added mechanism to ignore duplicate errors


MartinMystikJonas:
Still some error with Latte 2 I will check it.

src/Compiler/NodeVisitor/CopyDefinedVarsToOtherMethodsNodeVisitor.php
```diff
@@ -0,0 +1,108 @@
...
+        $statements = (array)$node->stmts;
+        $node->stmts = array_merge($this->definedVarsStatements, $statements);
+        return $node;
+    }
+
+    private function isEndOfTemplateHead(Node $statement): bool
+    {
+        if ($statement instanceof Echo_) {
+            // echo LR\Filters::escapeHtmlText($foo) or echo Nette\Bridges\FormsLatte\Runtime::renderFormBegin($foo) etc
+            if (($statement->exprs[0] ?? null) instanceof StaticCall) {
```
lulco: This condition isn't good as I see it now. What if we have some static call in “header” part and after it we assign some variable?

lulco:
I don’t think that removing duplicate errors is correct, we can hide some real case where there are really two errors with same message on same line. For example in forms I had:
`
{input something}{label something}
`
There are two errors that field `something` probably doesn't exist but it is correct they are two of them.

lulco:
> Still some error with Latte 2 I will check it.

Maybe some lost of types because of no expression holding doc types? That was always my biggest problem

src/Compiler/NodeVisitor/CopyDefinedVarsToOtherMethodsNodeVisitor.php
MartinMystikJonas: Yes this should be removed

MartinMystikJonas:
> I don’t think that removing duplicate errors is correct, we can hide some real case where there are really two errors with same message on same line. For example in forms I had: `{input something}{label something}` There are two errors that field `something` probably doesn't exist but it is correct they are two of them.

Is this really problem that it is not reported twice?

lulco:
> > I don’t think that removing duplicate errors is correct, we can hide some real case where there are really two errors with same message on same line. For example in forms I had: `{input something}{label something}` There are two errors that field `something` probably doesn't exist but it is correct they are two of them.
> 
> Is this really problem that it is not reported twice?

I don’t know, just thinking about it that it is problem because it hides real problem. It’s like using undefined variable twice but report it just once.

I think we should copy just what is needed to  be copied.

what about to not use marker but copy all assign nodes? Excluding known assingns like $_tmp (used for components) - here we just break iteration.

lulco:
Nevermind, functions like preg_match, preg_match_all can create variable without assinging… it will not work correctly, too.

MartinMystikJonas:
I think best would be to keep variable types surrounded by extract on both ends so it does not attach to random first stmt that follows. 

EDIT: It cause PHPStan errors in some cases (inside blocks). So what about some neutral expression like `extract([]);` or assignment to fictional variable?

MartinMystikJonas:
> I think we should copy just what is needed to be copied.

I am afraid there is no way how to know what needs to be copied and what not.

I think it is ok to have error reported just once even it is triggered multiple times on that line. You need to check that line anyways. And in case you fix only one cause then second cause will still be reported.

lulco:
> I think best would be to keep variable types surrounded by extract on both ends so it does not attach to random first stmt that follows.
> 
> EDIT: It cause PHPStan errors in some cases (inside blocks). So what about some neutral expression like `extract([]);` or assignment to fictional variable?

Assignment of fictional variable is my solution with marker… problem stays - we don’t know where to place this statement.

MartinMystikJonas:
> Assignment of fictional variable is my solution with marker… problem stays - we don’t know where to place this statement.

IMHO there is no way to prevent duplicate errors by copying incomplete header code. Duplicate errors will still stay in many common cases (use of undefined variable in header code). And any early terminating condition we can came up with could trim also important statements that follows.

So I would prefer to copy entire header code (in Latte 3 entire `prepare` except final return of defined vars, in Latte 2 up to confition that contains return of defined vars) and keep mechanism to ignore duplicate errors.

src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
```diff
@@ -67,7 +68,15 @@ public function enterNode(Node $node): ?Node
 
             $variableStatements[] = $docNop;
         }
-        $node->stmts = array_merge($variableStatements, $statements);
+
+        $variableStatements[] = new Expression(
```
lulco, line 72: This visitor is executed multiple times, so this is not very good way to mark end of variables. Also it is added to method before header assignements so it will not help

MartinMystikJonas:
Fixed for Latte 2 (reason for errors was that header code from `main` was copied into `prepare` method).

lulco:
> Fixed for Latte 2 (reason for errors was that header code from `main` was copied into `prepare` method).

That's correct, we have to copy vatiables to all other methods.

MartinMystikJonas:
I think we have only two options here now:
1) Errors in header code would be reported multiple times
2) All duplicate errors would be ignored

I would preffer option 2 - I think reporting given error on give line only once is enough even if it is triggered multiple times. You will have to check and fix that line anyways. And information that error is there multiple times is mostly useless.

MartinMystikJonas:
> > Fixed for Latte 2 (reason for errors was that header code from `main` was copied into `prepare` method).
> 
> That's correct, we have to copy vatiables to all other methods.

Not into prepare because prepare is not content renderting method in Latte 2. And would not have access to variables defined in header code inside `main` anyways - because `prepare` is called before `main`

lulco (commented):
Ok keep duplicate errors removing. Hope it will be good enough

MartinMystikJonas:
```php
	/**
	 * Renders template.
	 * @internal
	 */
	public function render(?string $block = null): void
	{
		$level = ob_get_level();
		try {
			$this->prepare();
			if (!$this->doRender($block)) {
				$this->main();
			}

		} catch (\Throwable $e) {
			while (ob_get_level() > $level) {
				ob_end_clean();
			}

			throw $e;
		}
	}
```

lulco:
> > > Fixed for Latte 2 (reason for errors was that header code from `main` was copied into `prepare` method).
> > 
> > 
> > That's correct, we have to copy vatiables to all other methods.
> 
> Not into prepare because prepare is not content renderting method in Latte 2. And would not have access to variables defined in header code inside `main` anyways - because `prepare` is called before `main`

Why it works for me before? I had some use case where parameters were used crossed methods. Just don’t remember which latte version and which method it was.

src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
MartinMystikJonas: That code will be in final template multiple times but it does nothing so I think it is not a problem.

Point of this code is that variable type doc comments are attached to this placeholder node. So it is copied even when next statement (return) is not. I think variables are first added to prepare/main and then it is copied from there together with header assignments.

lulco:
> I think we have only two options here now:
> 
> 1. Errors in header code would be reported multiple times
> 2. All duplicate errors would be ignored
> 
> I would preffer option 2 - I think reporting given error on give line only once is enough even if it is triggered multiple times. You will have to check and fix that line anyways. And information that error is there multiple times is mostly useless.

Prefered way is to not report duplicities and not ignore any error. Hardest, but the only correct, solution is to add some marker to latte itself. But it's little bit utopic.

MartinMystikJonas:
> > > > Fixed for Latte 2 (reason for errors was that header code from `main` was copied into `prepare` method).
> > > 
> > > 
> > > That's correct, we have to copy vatiables to all other methods.
> > 
> > 
> > Not into prepare because prepare is not content renderting method in Latte 2. And would not have access to variables defined in header code inside `main` anyways - because `prepare` is called before `main`
> 
> Why it works for me before? I had some use case where parameters were used crossed methods. Just don’t remember which latte version and which method it was.

It was wrong before but it does not manifest because content copied to `prepare` was in tests we used trimmed before code that coused error.

MartinMystikJonas:
> > I think we have only two options here now:
> > 
> > 1. Errors in header code would be reported multiple times
> > 2. All duplicate errors would be ignored
> > 
> > I would preffer option 2 - I think reporting given error on give line only once is enough even if it is triggered multiple times. You will have to check and fix that line anyways. And information that error is there multiple times is mostly useless.
> 
> Prefered way is to not report duplicities and not ignore any error. Hardest, but the only correct, solution is to add some marker to latte itself. But it's little bit utopic.

Only way I can think of is to mark every line in copied header code with `/** @phpstan-ignore-line */` so error would be reported only in original location.

src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
lulco: I think it is useless. What is the reason to add it here if it does nothing?

lulco:
Create this PR with my branch as target to see better diff, please

lulco:
> Create this PR with my branch as target to see better diff, please

Oh I can do it so I did :)

src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
MartinMystikJonas: It hoes nothing except that it binds type doc comments. We need to bind them to some neutral node instead of random first node that follows in method body. So it is not stripped during copying and also to prevent potential errors. I encountered strange error in my project that I solved by workaround but now I see it was caused by this.

That error manifests when first statement in method body is foreach. Doc comment above foreach is allowed to define only key and value avriables of foreach otherwise PHPStan reports error.

MartinMystikJonas: I think I should add test for that now

lulco:
> > > I think we have only two options here now:
> > > 
> > > 1. Errors in header code would be reported multiple times
> > > 2. All duplicate errors would be ignored
> > > 
> > > I would preffer option 2 - I think reporting given error on give line only once is enough even if it is triggered multiple times. You will have to check and fix that line anyways. And information that error is there multiple times is mostly useless.
> > 
> > 
> > Prefered way is to not report duplicities and not ignore any error. Hardest, but the only correct, solution is to add some marker to latte itself. But it's little bit utopic.
> 
> Only way I can think of is to mark every line in copied header code with `/** @phpstan-ignore-line */` so error would be reported only in original location.

“Line” in php code is not the line in latte. In compiled code it can be multiple lines so only the first line error will be ignored probably. I'm already reconciled with removing duplicate errors in error builder ;)

src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
lulco: Ah it's “doc keeper” statement like the `$__marker___ = $this;` was in my PR. Got it now

lulco (approved):

src/Compiler/NodeVisitor/AddVarTypesNodeVisitor.php
MartinMystikJonas: Test added

MartinMystikJonas:
If you are ok with this you can merge it into your branch and then merge your PR.

lulco:
> If you are ok with this you can merge it into your branch and then merge your PR.

Yes merge it please and I’ll merge my

lulco:
By the way, now I recall that var comments are not copied in actual version, they are added to all methods separatelly. The main reason are Block methods where we have to check if the variable is also in method's arguments and if so, skip it. So adding assignment has no reason imho

MartinMystikJonas:
It is still needed to avoid that potential phostan errors caused by binding them to first stmt of method body i mentioned.

PR #187, opened 2023-01-04
Set of small fixes I found in real world testing

MartinMystikJonas:


src/Rule/LatteTemplatesRule.php
```diff
@@ -98,6 +98,10 @@ public function processNode(Node $collectedDataNode, Scope $scope): array
             }
         }
 
+        if (count($errors) > 1000) {
+            $errors[] = RuleErrorBuilder::message('Too many errors in latte.')->build();
+        }
```
lulco, line 103: Here should be some break / return

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -170,7 +170,7 @@ public function afterTraverse(array $nodes): array
                     new Return_(new Variable('control')),
                 ])
                 ->makePublic()
-                ->setReturnType($componentType);
+                ->setReturnType('Nette\ComponentModel\IComponent');
```
lulco, line 173: 👍

src/Compiler/TypeToPhpDoc.php
```diff
@@ -21,6 +22,10 @@ public function __construct(TypeStringResolver $typeStringResolver)
 
     public function toPhpDocString(Type $type): string
     {
```
lulco, line 24: What does this? When it appears? Can we add some test?

lulco (approved):

src/Rule/LatteTemplatesRule.php
MartinMystikJonas: This is after analysis is done. I break analysis in different palce. BUt to prevent adding this error multiple times (for every resolver and resolved node) I add it at the end or analysis

src/Compiler/TypeToPhpDoc.php
MartinMystikJonas: When you assign $this to variable it generated invalid phpdoc because that type is outputted as `$this(\Some\Class)`

MartinMystikJonas: I necountered it in our code where we passed $this to be able to call method form $this in template

lulco: I saw that before, but somehow I always change the code to get Some\Class. So can you add test case?

PR #189, opened 2023-01-04
Prevent duplicate analysis of same template with same context

MartinMystikJonas:


src/LatteTemplateResolver/LatteTemplateResolverResult.php
```diff
@@ -36,7 +36,7 @@ public function __construct(array $templates = [], array $errors = [])
     }
 
   /**
-   * @return Template[]
+   * @return array<string, Template>
```
lulco, line 39: Where is the string key set? What it is? Some hash?

MartinMystikJonas: Yes it is hash of template with context. It is set in addTemplate already to prevent duplicate template in result. But it did not work across results from different resolvers. So I made it part of interface and I am merging tempmates in rule

lulco: I see. OK. So we are not going to debug why it acts weird when the context is same for more runs?

lulco (approved):

src/LatteTemplateResolver/LatteTemplateResolverResult.php
MartinMystikJonas: Well we do not want duplicate analysis anyways (all error would be doubled) so I see no point in debugging it. ANd I am afrait it would be some internal state of PHPStan that is in play here.

lulco: So if one template is rendered from more actions / presenters with same variables, components etc. it will be analysed just once and also reported only once for the first occurence.

Then I fix this error in one presenter and I will think it's already fixed. But it will not be fixed, it will  just appear in other presenter.

I think we should rather use presenter / control name in the context hash. Analyse template once is cool but we need to print all errors imho

MartinMystikJonas: Presenter/control class and action is part of that hash already.

MartinMystikJonas:
```php
    #[ReturnTypeWillChange]
    public function jsonSerialize()
    {
        return [
            'path' => $this->path,
            'actualClass' => $this->actualClass,
            'actualAction' => $this->actualAction,
            'variables' => $this->variables,
            'components' => $this->components,
            'filters' => $this->filters,
            'parentTemplatePaths' => $this->parentTemplatePaths,
        ];
    }
```

MartinMystikJonas: Hash is created from json_serialized representation of Template that includes actualClass, actualAction full chain of includes, ...

MartinMystikJonas: Point it to analyse template in all contexts but in every context just once.

lulco: Then how it is possible it creates the same hash two times? Some custom resolver finds the same template path?

lulco:
ok, merge both PR, I have some time for real world testing, so I will go with main

src/LatteTemplateResolver/LatteTemplateResolverResult.php
MartinMystikJonas: Yes in my case it was caused by custom resolver sometimes resolving it exactly the same as default resolver

PR #190, opened 2023-01-04
Fixes

lulco:
- switched AddVarTypesNodeVisitor for global variables and local variables
- reseted forms in FormsNodeVisitorBehavior
- changed condition for errorFieldNodes (find Stmt)

### DIFF
--- a/src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
@@ -23,9 +23,8 @@ use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use PhpParser\Node\Scalar\String_;
-use PhpParser\Node\Stmt\Echo_;
+use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Expression;
-use PhpParser\Node\Stmt\If_;
 use PhpParser\Node\Stmt\Return_;
 use PhpParser\NodeVisitorAbstract;
 use PHPStan\Type\VerbosityLevel;
@@ -108,11 +107,11 @@ final class AddFormClassesNodeVisitor extends NodeVisitorAbstract implements For
                         break;
                     }
 
-                    $rootParentNode = $parentNode;
-
-                    if ($rootParentNode instanceof Echo_ || $rootParentNode instanceof If_) {
+                    if ($parentNode instanceof ClassMethod) {
                         break;
                     }
+
+                    $rootParentNode = $parentNode;
                 }
                 $this->errorFieldNodes[] = [
                     'node' => $rootParentNode,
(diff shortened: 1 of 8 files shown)


src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -183,4 +211,62 @@ public function afterTraverse(array $nodes): array
...
+
+        if (!$expr->var instanceof PropertyFetch) {
+            return false;
+        }
+
+        if (!$expr->var->name instanceof Identifier) {
+            return false;
+        }
+
+        if ($expr->var->name->name !== 'global') {
```
lulco: We should use NameResolver

src/Error/ErrorBuilder.php
```diff
@@ -26,6 +26,7 @@ final class ErrorBuilder
         '/Parameter #1 \$blocks of method Nette\\\\Bridges\\\\ApplicationLatte\\\\SnippetDriver::renderSnippets\(\) expects .*/', # nette internal error
         '/Parameter #3 \$blocks of static method Nette\\\\Bridges\\\\ApplicationLatte\\\\UIRuntime::initialize\(\) expects .*/', # nette internal error
         '/Cannot call method getParent\(\) on Latte\\\\Essential\\\\CachingIterator\|null\./', # latte internal error
+        '/Cannot call method attributes\(\) on Nette\\\\Utils\\\\Html\|null\./', # latte internal error
```
lulco, line 29: I hope this will not hide some real errors

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -64,15 +70,31 @@ public function enterNode(Node $node): ?Node
             }
             $dimFetch = $assign->expr;
 
+            if (!$this->isUiControl($dimFetch->var)) {
```
lulco, line 73: Why not just check if it is formStack? It can be anything else, why checking uiControl?

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -64,15 +70,31 @@ public function enterNode(Node $node): ?Node
...
+
             $formNameString = $dimFetch->dim;
             $formName = $formNameString->value;
 
             $formClassName = $this->formClassNames[$formName] ?? null;
             if ($formClassName === null) {
-                return null;
+                $this->actualForm = new Form($formName, new ObjectType('Nette\Forms\Form'));
+                $error = new Error('Form with name "' . $formName . '" probably does not exist.');
+                return new Assign(
```
lulco, line 88: This is kind of hack :))

MartinMystikJonas: I took me a lot of thinking but it works :-D

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
MartinMystikJonas: in this case we are looking for `$form = $this->global->formsStack[] = $this->global->uiControl[''formName']` and we are on `uiControl['formName']` node

src/Error/ErrorBuilder.php
MartinMystikJonas: Well it might but I do not think it is probable. You usually do not manipulate Html objects in latte tempalte directly

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
MartinMystikJonas: ok

PR #199, opened 2023-01-06
Dynamic class

MartinMystikJonas:
RelatedFilesResolver will fail when it found `new $class()`

### DIFF
--- a/src/Resolver/ValueResolver/ValueResolver.php
+++ b/src/Resolver/ValueResolver/ValueResolver.php
@@ -104,4 +104,16 @@ final class ValueResolver
             return null;
         }
     }
+
+    /**
+     * @return string[]|null
+     */
+    public function resolveStrings(Expr $expr, Scope $scope): ?array
+    {
+        $values = $this->resolve($expr, $scope);
+        if ($values === null) {
+            return null;
+        }
+        return array_filter($values, 'is_string');
+    }
 }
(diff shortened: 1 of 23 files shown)


tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/ResolvePresenter.php
```diff
@@ -94,4 +94,10 @@ public function actionSendTemplateUnresolvable(): void
         $this->template->send = 'send';
         $this->sendTemplate(new DefaultTemplate(new Engine()));
     }
+
+    public function dynamicClassName()
+    {
+        $class = uniqid();
```
lulco, line 100: What about some string representing existing class name? It should work.

MartinMystikJonas: I think we could use value resolver if it is expression

lulco: You have scope here, isn't it better?

MartinMystikJonas: Value resolver uses scope internally in cases where it is useful

lulco: 👍

lulco (approved):

MartinMystikJonas:
I added resolveStrings to ValueResolver because in almost all palces we need just strings and we were filtering non+string values out repeatedly. Ok?

src/Resolver/ValueResolver/ValueResolver.php
```diff
@@ -104,4 +104,16 @@ public function resolve(Expr $expr, Scope $scope, $unknownValuePlaceholder = nul
             return null;
         }
     }
+
+    /**
+     * @return string[]|null
+     */
+    public function resolveStrings(Expr $expr, Scope $scope)
```
lulco: Add return type ?array

lulco (approved):

PR #204, opened 2023-01-06
Standalone resolver

MartinMystikJonas:


src/Collector/Finder/ResolvedNodeFinder.php
```diff
@@ -18,15 +21,43 @@ final class ResolvedNodeFinder
...
+     * @param string[] $existingTemplates
+     * @param LatteTemplateResolverInterface[] $latteTemplateResolvers
+     */
+    public function __construct(CollectedDataNode $collectedDataNode, array $existingTemplates, array $latteTemplateResolvers)
     {
+        // node resolvers
         $collectedResolvedNodes = ResolvedNodeCollector::loadData($collectedDataNode, CollectedResolvedNode::class);
+
+        // file resolvers
+        foreach ($existingTemplates as $templateFile) {
```
lulco: I think we don't need this part, only LatteCustomTemplateResolverInterface. If we want to use existing templates we can implement it with LatteCustomTemplateResolverInterface. It solves my and also mabar's problem.

MartinMystikJonas: Well I added it first and I think it might be most common use case. Also it reduces dependencies on kinf-of internal classes like AnalasedTemplatesRegistry - used to find latte files

MartinMystikJonas: But I can remove it and keep jsut universal custom resolver

lulco: if I'm right then yes please, remove it.

lulco:
I have one question here :) how will I say to this custom collected nodes how they should find variables, components and templates?

MartinMystikJonas:
You mean what files should be collected for latte context?

Second parameter of CollectesResokvedNode is analysedFile where collecting should start. 

If you mean how to find assigned variables from some method for exame then use is same as with normal resolvers. You ask VariablesFinder for variables in given class method.

src/Collector/Finder/ResolvedNodeFinder.php
MartinMystikJonas: REmoved. We can make that code part of some kind of AbstractFileResolver later

lulco:
VariableFinder is closed now, so external developers can’t change it in their extensions where they wrote some custom resolvers. I still didn’t get to try custom resolvers in my projects, so this is just my actual opinion. Also I am not sure if I should to try it if you’re refactoring it again in next pr :)

src/Collector/CollectedData/CollectedResolvedNode.php
```diff
@@ -7,21 +7,24 @@
 use PHPStan\ShouldNotHappenException;
 
 /**
- * @phpstan-type CollectedResolvedNodeArray array{resolver: string, params: array<string, string>}
+ * @phpstan-type CollectedResolvedNodeArray array{resolver: string, analysedFile: string, params: array<string, string>}
```
lulco, line 10: This was added because of first custom resolver so it's not needed for actual state

lulco:
Please remove all unnecessary changes which were added with first draft. Keep just code for “open” custom resolver

src/Collector/CollectedData/CollectedResolvedNode.php
MartinMystikJonas: No it is needed with custom resolvwrs too to let analyser know where to collect context. For example if tempalte belongs to some control in vendor you need to pass file with that control. Otherwise it would not be collected for latte context

MartinMystikJonas:
> VariableFinder is closed now, so external developers can’t change it in their extensions where they wrote some custom resolvers. I still didn’t get to try custom resolvers in my projects, so this is just my actual opinion. Also I am not sure if I should to try it if you’re refactoring it again in next pr :)

You do not need to change it. You jsut need to use it.

src/Collector/CollectedData/CollectedResolvedNode.php
MartinMystikJonas: In Node collectors. File where node is located is root of latte context analysis - and we extracted it from array keys used by PHPStan in CollectedDataNode. BUt for custom resolvers we need way to say it explicitly to be able to collect context from files not matched by any node-based collector.

lulco:
I hope I’ll check it today and create some issue if it will not work. Thanks for now. Let’s merge it. We will see.

src/Rule/LatteTemplatesRule.php
```diff
@@ -82,11 +81,8 @@ public function getNodeType(): string
      */
     public function processNode(Node $collectedDataNode, Scope $scope): array
     {
-        $resolvedNodes = $collectedDataNode->get(ResolvedNodeCollector::class);
-        $resolvedNodeFinder = new ResolvedNodeFinder($collectedDataNode);
-
-        $processedFiles = array_unique(array_keys($resolvedNodes));
```
MartinMystikJonas, line 88: Originally we got processed files here. BUt with custom resolvers we need way to pass it explvitly

lulco (approved):

src/Rule/LatteTemplatesRule.php
lulco: I see, so it is passed same way from all resolvers - new and old. Ok

MartinMystikJonas:
If you are going to write custom resolvers I would first wait for #206. It makes some changes to how resolvers are written and you would need to rewrite it - nothing significant - basically only change of paraemter type in signature and different way to call finders.

lulco:
Ok I’ll wait

PR #207, opened 2023-01-07
Added support for collecting variables from $template->setParameters()

lulco:


src/Resolver/TypeResolver/TemplateTypeResolver.php
```diff
@@ -23,4 +28,23 @@ public function resolve(Type $type): bool
         }
         return false;
     }
+
+    public function resolveByNodeAndScope(Node $node, Scope $scope): bool
```
MartinMystikJonas, line 32: Maybe some nicer names? Like resolveNode and resolveType?

MartinMystikJonas (approved):

src/Resolver/TypeResolver/TemplateTypeResolver.php
lulco: Well its always resolving if type is template. Maybe I’ll use resolveByType instead of resolve. I’ll think about it later when I’ll use it

PR #211, opened 2023-01-09
Annotations for all variable sub-collectors

### LINKED ISSUE #208: Add phpdoc handling for multi variables assigns
It should merge collected types with phpdoc types
examples:
```php
[$this->template->foo, $this->template->bar] = ['foo', $this->bar()];
$this->template->setParameters([
    'baz' => 'baz',
    'qwerty' => $this->qwerty(),
]);
```
bar() and qwerty() has unknown types, and we want to set them via annotations.

MartinMystikJonas:
Resolves #208 

I unified handling of `@phpstan-latte-var` for single and multi assigns. There is change of behaviour in one edge case. When annotation is used with explicit name above assign with different name both variables (assigned and from phpdoc) are collected.

I has to chenge interface of subcollector to differentiate between ""this is not variable assign" (null) and "this is variable assign but I do not know what is assigned" (empty array). Becaue annotation processing have to behave differently in those cases. For not assignment nodes whay must be ignored (because they might belong to another type of node like render call)

I also fixed bug in NAmeResolver. Previous version would resolve `$this->tempalte->{$varName}` as assignment to `varName` property same as `$this->template->varName`


src/Resolver/NameResolver/NameResolver.php
```diff
@@ -31,10 +32,10 @@ public function resolve($node): ?string
         if ($node instanceof Name || $node instanceof Identifier) {
             return $this->resolve((string)$node);
         }
-        if ($node instanceof Variable) {
+        if ($node instanceof Variable && !$node->name instanceof Expr) {
```
lulco: why this condition here? if you pass it to resolve method, it will evaluated to null because no if will match

src/LatteContext/Collector/VariableCollector/AssignToArrayOfTemplateVariablesCollector.php
```diff
@@ -13,12 +16,32 @@
...
+    }
+
+    /**
+     * @param Assign $node
+     */
+    public function collect(Node $node, Scope $scope): ?array
     {
         if (!($node->var instanceof Array_ || $node->var instanceof List_)) {
-            return [];
+            return null;
```
lulco, line 44: why null? we have to handle null specially, can't merge it with all other arrays

src/LatteContext/Collector/VariableCollector.php
```diff
@@ -51,13 +56,37 @@ public function collectData(Node $node, Scope $scope): ?array
...
                 continue;
             }
-            $collectedVariables[] = $variableCollector->collect($node, $scope);
+            $variables = $variableCollector->collect($node, $scope);
+            if ($variables === null) {
+                continue;
+            }
+            $isVariablesNode = true;
+            foreach ($variables as $variable) {
+                $collectedVariables[$variable->getVariableName()] = $variable;
```
lulco: this will override existing collected variables, it is correct?

MartinMystikJonas: It should not happen that multiple collectors match same node but if it happens I think metge is ok. But union would be better probably.

src/LatteContext/Collector/VariableCollector/AssignToArrayOfTemplateVariablesCollector.php
MartinMystikJonas: See PR comment. We need to differentiate between "not a variable node" and "it is a variable node but no variables could be resolved". Otherwise it conflicts with other colectors.

src/Resolver/NameResolver/NameResolver.php
MartinMystikJonas: `$$varName` or `${$varName}` will otherwise resolve as `$varName` not as unknown

src/LatteContext/Collector/VariableCollector.php
lulco: I can write any collector for example for Class_ node and then iterate over stmts and find some variables (just example).

MartinMystikJonas: Yes but would you write multiple collectors for Class_ that would resolve conflicting types for same variable name?

src/LatteContext/Collector/VariableCollector/AssignToArrayOfTemplateVariablesCollector.php
lulco: I see, it just add some complexicity to the code - we have to check null everywhere where it is used. Also I'm not sure if it is replaced everywhere it was needed.

src/LatteContext/Collector/VariableCollector.php
lulco: I don't need multiple collectors, My one collector can collect variables of same name multiple times, e.g. in case of I collect something like this with ClassMethod Node:
```
if ($condition) {
    $this->template->a = 'a';
} else {
    $this->template->a = 'b';
}
```
Variable `a` will be represented just with value `'b'` and not `'a'|'b'`

src/LatteContext/Collector/VariableCollector/AssignToArrayOfTemplateVariablesCollector.php
MartinMystikJonas: It is processed only in VariableCollector when using data from sub-collectors.

src/Resolver/NameResolver/NameResolver.php
lulco: isn't it case in all condition blocks?

src/LatteContext/Collector/VariableCollector/AssignToArrayOfTemplateVariablesCollector.php
lulco: Luckily :)

lulco (approved):

src/LatteContext/Collector/VariableCollector.php
MartinMystikJonas: I see. Ok I will change it to  union

src/Resolver/NameResolver/NameResolver.php
MartinMystikJonas: What do you mean by condition blocks? BUt theoretically it can happend enywhere variable-variable can occur. So even a method name

lulco: I thought in each if in this class :) so I think solution should be add if ($node instanceof Expr) { return null; } at the beginging of resolve method or after 
```
    if ($node === null) {
            return null;
        }
        if (is_string($node)) {
            return $node !== '' ? $node : null;
        }
```

MartinMystikJonas: That would make resolving Variable impossible because Variable is Expr.

MartinMystikJonas: We need to limit resolving of Expr(Variable) to only one level but do not allow to pass it to recursive resolve call

MartinMystikJonas: So this condition should be for every call to $this->resolve in this class

MartinMystikJonas: Or maybe add something like `resolveNext` that would contain this safeguard and then call resolve. And in all branches we should call only resolveNext

lulco: resolveNext could be like this:
```
private function resolveNext(Node $node)
{
if ($node instanceof Expr) {
   return null;
}
return $this->resolve($node);
```

right?

MartinMystikJonas: Yes exactly

PR #213, opened 2023-01-09
Improving finders with multiple method names in one call

lulco:
This begins as hot fix for collecting variables from methods which are overrided in traits (or parents), but it is also optimisation because some methods are called just once e.g. 
```
$this->findInClasses($className)
```
and
```
$this->findInMethodCalls($className, '__construct')
```
in VariableFinder from NetteApplicationUIPresenterGlobals

### DIFF
--- a/src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
+++ b/src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
@@ -13,8 +13,7 @@ trait NetteApplicationUIPresenterGlobals
         $className = $reflectionClass->getName();
         $presenterType = new ObjectType($reflectionClass->getName());
         return array_merge(
-            $this->variableFinder->find($className, 'startup'),
-            $this->variableFinder->find($className, 'beforeRender'),
+            $this->variableFinder->find($className, 'startup', 'beforeRender'),
             [
                 new Variable('presenter', $presenterType),
                 new Variable('control', $presenterType),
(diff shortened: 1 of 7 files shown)


src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
```diff
@@ -25,30 +24,18 @@ protected function getClassGlobalVariables(ReflectionClass $reflectionClass): ar
     protected function getClassGlobalComponents(ReflectionClass $reflectionClass): array
     {
         $className = $reflectionClass->getName();
-        return array_merge(
-            $this->componentFinder->find($className, ''),
```
lulco, line 29: do you remember why we had `''` here? :) I've removed it and it still works

src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
```diff
@@ -25,30 +24,18 @@ protected function getClassGlobalVariables(ReflectionClass $reflectionClass): ar
...
-            $this->componentFinder->find($className, 'beforeRender')
-        );
+        return $this->componentFinder->find($className, 'startup', 'beforeRender');
     }
 
     protected function getClassGlobalForms(ReflectionClass $reflectionClass): array
     {
         $className = $reflectionClass->getName();
-        return array_merge(
-            $this->formFinder->find($className, ''),
```
lulco, line 39: Same here

src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
```diff
@@ -25,30 +24,18 @@ protected function getClassGlobalVariables(ReflectionClass $reflectionClass): ar
...
-            $this->formFinder->find($className, 'beforeRender')
-        );
+        return $this->formFinder->find($className, 'startup', 'beforeRender');
     }
 
     protected function getClassGlobalFilters(ReflectionClass $reflectionClass): array
     {
         $className = $reflectionClass->getName();
-        return array_merge(
-            $this->filterFinder->find($className, ''),
```
lulco, line 49: and here

src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
MartinMystikJonas: Components can be declared globally for class (not only dor specific method). But I think we include global components to every method. So as long as you finding for at least one method globaln components are included and this is redundant.

MartinMystikJonas (approved):

src/LatteTemplateResolver/NetteApplicationUIPresenterGlobals.php
lulco: that is collected via $this->findInClasses($className) so methods are not required

PR #214, opened 2023-01-09
Overriding method in trait

lulco:
Failing test case for overriding method in trait.
I figure out that we also don't have support for setFile() with method call inside.

### DIFF
--- a/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/TraitOverriding/BaseControl.php
+++ b/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/TraitOverriding/BaseControl.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\TraitOverriding;
+
+use Nette\Application\UI\Control;
+
+abstract class BaseControl extends Control
+{
+    protected function getTemplatePath(): string
+    {
+        return __DIR__ . '/base.latte';
+    }
+}
--- a/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/TraitOverriding/SomeControl.php
+++ b/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/TraitOverriding/SomeControl.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\TraitOverriding;
+
+final class SomeControl extends BaseControl
+{
+    public function render(): void
+    {
+        $this->template->a = 'a';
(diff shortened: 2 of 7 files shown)


MartinMystikJonas:
Support for method calls resolved as value are limitation of PHPStan. BEcause PHPStan knows only method signature not a method body at point wheere we needs to evaluate method call. I think we could solve it by collector somehow but it would not be easy. BEcause when we ask about that method we knows it only returns string but how exact value.

lulco:
🤔 can we somehow collect returned value via ValueResolver and use it as type of this method call node?

MartinMystikJonas:
We can call return values from return statements. In fact we do basically the same thing for components already. But problem is order of analysis. We might collect calling class before called class so no data would be collected at that point. So we would need two stage collection. In first step we would collect return types and related files. And in second step (when resturn types are known) we can collect render calls, etc. using this information to evaluate values.

MartinMystikJonas:
It would be possible but not easy and could greatly influnce performance. In our code where this is used I solved it either by custom tempalte resolver or `@phpstan-latte-tempalte` annotation

lulco:
Can I create my own collector where I will collect it and it will get to Latte Context with other finders and I will use it in my Custom Resolver?

lulco:
> It would be possible but not easy and could greatly influnce performance. In our code where this is used I solved it either by custom tempalte resolver or `@phpstan-latte-tempalte` annotation

I have thousands such components, I don't want to use annotations for all. or is it possible to say to render() something like "hey, look at method getTemplatePath, it is your file you will render"

MartinMystikJonas:
@lulco There is problem that expressions must be evaluted in collectors where you have actual Scope. At any higher level (finder/tempalte resolver) you know only result from collector and you cannot make additional Expr evaluation because you no longer have scope in correct state from point where expression is written.

MartinMystikJonas:
You can write collector that would collect vale returned from that method. Then you can access it in custom template resolver.

MartinMystikJonas:
Or maybe you can collect it as `CollectedTemplatePath` attached to given class (not method)

MartinMystikJonas:
But right now I think you will get error about unresolvable render call because we cannot evaluate what is passed to render method. We look up CollectedTemplatePath only if parameter is missing or null is passed.

lulco:
> You can write collector that would collect vale returned from that method. Then you can access it in custom template resolver.

How can I access it? In LatteContext there are only finders from phpstan-latte.

Nevermind, we can discuss this in wednesday on call. I'll prepare more questions.

MartinMystikJonas:
You would have to collect it either as CollectedTemplatePath or as CollectedMethodCall of custom type (see how setView is handled)

MartinMystikJonas:
I am thinking about it little more and I think for simple expressions we can do evaluation in finders. Similar to how we handle wildcards. We will resolve method call as some placeholder with class+method. And we will replace placeholders with collected values in finder.

MartinMystikJonas:
I think I solved it. I will prepare PR, but it seems to work just fine.

MartinMystikJonas:
Solved in #217

PR #216, opened 2023-01-09
Fixed method call traversing for complex hierarchies

MartinMystikJonas:


MartinMystikJonas:
Your problem with traits was only about `getTempaltePath` or was it with other emethod calls tooo? Can you test this branch if it is resolved? If not please create additional tests cases in SomeControl/Hierarchy

tests/Rule/LatteTemplatesRule/SimpleControl/config.neon
```diff
@@ -0,0 +1,4 @@
+parameters:
+    latte:
+        collectedPaths:
```
lulco, line 3: You should create parent and grandparent in source folder and not collect data from whole fixtures directory

lulco: Or run analyse on 3 files instead of one

lulco:
I think it was only about the method getTemplatePath. I just get some weird (for me) errors about template file not set for render() or something like that. Then I realize  I'm using trait. Maybe it wasn’t about trait but also about resolving value of that method

lulco:
I have branch latte-context installed so I can't install this one too :(

MartinMystikJonas:
Ok I will finish this and resolving of paths with method calls and then I will rebace latte-context ontho main

tests/Rule/LatteTemplatesRule/SimpleControl/config.neon
MartinMystikJonas: It does not work because analyse is called on every file seaprately in tests. And I did not wanted to move it to Source when it is just for single fixture.

It is done same way in PresenterWithoutMOdule

MartinMystikJonas: Can I keep it this way or do you want me to move it to source even for Fixture specific code?

lulco: keep it

MartinMystikJonas:
Ok so can I merge it?

lulco (approved):

PR #218, opened 2023-01-10
Added try catch to ConstExprEvaluator in LinkNodeVisitor

lulco:


src/Compiler/NodeVisitor/LinkNodeVisitor.php
```diff
@@ -88,7 +89,11 @@ private function prepareNodes(MethodCall $methodCall, array $attributes): ?array
         $linkArgs = $methodCall->getArgs();
         $target = $linkArgs[0]->value;
 
-        $targetName = (new ConstExprEvaluator())->evaluateDirectly($target);
+        try {
+            $targetName = (new ConstExprEvaluator())->evaluateDirectly($target);
+        } catch (ConstExprEvaluationException $e) {
+            $targetName = null;
```
MartinMystikJonas: maybe just return null?

MartinMystikJonas (approved):

src/Compiler/NodeVisitor/LinkNodeVisitor.php
lulco: You're right, we can do that :)

PR #221, opened 2023-01-10
Template path subcollectors

lulco:
In my apps I need to collect template paths specially. It is just return in some method, e.g.:
```php
protected function templateFilePath(): string
{
    return __DIR__ . '/default.latte';
}
```

### DIFF
--- a/src/LatteContext/Collector/TemplatePathCollector/TemplatePathCollectorInterface.php
+++ b/src/LatteContext/Collector/TemplatePathCollector/TemplatePathCollectorInterface.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\LatteContext\Collector\TemplatePathCollector;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+
+interface TemplatePathCollectorInterface
+{
+    public function isSupported(Node $node): bool;
+
+    /**
+     * @return string[]
+     */
+    public function collect(Node $node, Scope $scope): array;
+}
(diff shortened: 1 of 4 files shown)


src/LatteContext/Collector/TemplatePathCollector/TemplatePathCollectorInterface.php
```diff
@@ -0,0 +1,18 @@
...
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+
+interface TemplatePathCollectorInterface
+{
+    public function isSupported(Node $node): bool;
+
+    /**
+     * @return string[]|null
```
MartinMystikJonas, line 15: Why string and not CollectedTemplatePath? In VariableSubCollectors we also return CollectedVariable. Would not be it better to keep it similar?

lulco: Because you have there special case for null. If any collector find template in node, we have to return it. If CollectedTemplatePath is returned here, we have no chance to change it. So we would have errors which we don’t want.

MartinMystikJonas: make sense. I jsut wonder if it would not be better for variable sub collectors also return just name=>type instead of collectedVariable

MartinMystikJonas (approved):

src/LatteContext/Collector/TemplatePathCollector/TemplatePathCollectorInterface.php
lulco: Maybe, in my PR it was just simple add collected variable to array. Then it becames more complicated. I'm not sure what is better for that use case. But I don’t very like some array shaped returns (name, type). CollectedVariable seems nice to me.

PR #223, opened 2023-01-11
Collect FormField name default value

MartinMystikJonas:


src/LatteContext/Collector/FormFieldCollector.php
```diff
@@ -103,6 +105,12 @@ public function collectData(Node $node, Scope $scope): ?array
             $formFieldType = $formFieldParametersAcceptor->getReturnType();
 
             $fieldNameArg = $node->getArgs()[0] ?? null;
+            $fieldNameDefaultType = $formFieldParametersAcceptor->getParameters()[0]->getDefaultValue();
+            if ($fieldNameDefaultType instanceof ConstantStringType) {
+                $fieldNameDefault = trim($fieldNameDefaultType->getValue(), '"\'');
```
MartinMystikJonas, line 110: For some reason PHPStan returns this including "" or '' but that are invalid characters for form field anyway so I trimed it

lulco: Yeah I did it too somewhere :)

lulco (approved):
that was quick :) LGTM

PR #224, opened 2023-01-11
Fix: support for union type for custom form methods

lulco:
Follow up for https://github.com/efabrica-team/phpstan-latte/pull/223

I had to fix some issues

### DIFF
--- a/src/LatteContext/Collector/FormFieldCollector.php
+++ b/src/LatteContext/Collector/FormFieldCollector.php
@@ -15,6 +15,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\UnionType;
 
 /**
  * @extends AbstractLatteContextCollector<Node, CollectedFormField>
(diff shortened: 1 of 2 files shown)


src/LatteContext/Collector/FormFieldCollector.php
```diff
@@ -103,26 +104,39 @@ public function collectData(Node $node, Scope $scope): ?array
             }
 
             $formFieldType = $formFieldParametersAcceptor->getReturnType();
-
             $fieldNameArg = $node->getArgs()[0] ?? null;
-            $fieldNameDefaultType = $formFieldParametersAcceptor->getParameters()[0]->getDefaultValue();
+
+            $formFieldParameters = $formFieldParametersAcceptor->getParameters();
+            $fieldNameDefaultType = isset($formFieldParameters[0]) ? $formFieldParameters[0]->getDefaultValue() : null;
```
lulco, line 110: here I got error "Call method getDefaultValue() on null"

src/LatteContext/Collector/FormFieldCollector.php
```diff
@@ -103,26 +104,39 @@ public function collectData(Node $node, Scope $scope): ?array
...
             if ($fieldNameDefaultType instanceof ConstantStringType) {
                 $fieldNameDefault = trim($fieldNameDefaultType->getValue(), '"\'');
             } else {
                 $fieldNameDefault = null;
             }
         }
 
-        if (!$formFieldType instanceof ObjectType) {
-            return null;
+        $formFieldTypes = [];
```
lulco, line 118: My method addAccount() has union return type, so I'm trying to solve it here. It works like expceted, I just don't like the code. Any idea how to write it more phpstan-like?

MartinMystikJonas: Try TypeCombinator::intersect return type with union of allowed types.

MartinMystikJonas (approved):

src/LatteContext/Collector/FormFieldCollector.php
lulco: I'll play with it later, now I'm going to merge

PR #228, opened 2023-01-11
Fixed type of $presenter in Control templates

MartinMystikJonas:


lulco (approved):
At the begining it was this, where did it changed? :)

MartinMystikJonas:
No idea but it was there for longer time. Maybe my first resolvers rework

PR #232, opened 2023-01-12
export-ignore unneeded files

spaze:
These files are not needed for the operation of the extension, so they don't need to be in the distribution package. Removing these cuts the total number of files to half.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+.gitattributes export-ignore
+.github export-ignore
+.gitignore export-ignore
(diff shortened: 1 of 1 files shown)


lulco:
Thanks! Can you add docs too?

spaze:
Haven't `export-ignore`'d docs because I think they're useful when *using* the extension, not just when *building* it. Like let's say a README file. What do you think?

lulco:
Mmm Ok then.

Issue #163, opened 2022-12-20
Refactoring ideas


MartinMystikJonas:
- [x] Create `Form` and `FormField` value objects in `Tempalte` namespace - do not use `Collected*` versions in template/compiler (Collected versions are between collector and finder, normal versions are between finder and template/compiler)
- [x] Create `TemplateContext` value object that would group variables, components, forms and filters passed to tempalte in single object
- [x] ~~`Template` value object might be renamed to `TemplateRender` becaause it prepresents specific render of template~~
- [x] `Tempalte` object should internally use new `TemplateContext`
- [x] ~~Create `TemplateContextFinder` thet would return `TempalteContext` for given class/method - then resolvers would use mainly this finder and would not need to change whenever we add something new to `TempalteContext` (like we did with Forms and Filters recently)~~
- [ ] Remove `VariableCollector` namespace completely
  * `ʟ_*` variables should be handled by `Compiler`
  * `presenter`, `control`, `flashes` should be handled by `LatteTempalteResolver` (on things that actually sets them like presenters/controls)
  * not sure about `baseUrl` and `basePath` - we should investigate where they are set in Latte
  * dynamic filter variables should be handled by `Compiler`
  * globalTemplateVariables could be dedicated NodeVisitor that takes them as parameter
  * Maybe create something like `AbstractVarTypeNodeVisitor` that would contain logic of adding variables. New variables then could be added by creating additional NodeVisitors
- [x] Create `ComponentsHelper` with merge/union logis similar to `VariablesHelper` to properly handle conflicting definitions of component with same name (for example in parent and child)
- [x] Maybe change latte context resolvers to support multiple nodes at once like rector? To avoid unnecessary calls on irrelevant nodes.
- [x] Create LatteContextCollectorRegistry top handle logic of searching for relevant collectors for given node. Similar to PHPStan\Collectors\Registry
- [ ] (???) Create CollectLocaltion value object to hold class+method where thing was colected
- [ ] (???) Maybe create common ancestor for value objects that have class/method context to avoid duplication

PR #234, opened 2023-01-12
TemplateContext and unified Template items

MartinMystikJonas:
Draft implementation of ideas from #163 

Benefits:
- duplicated code from Helpers replaced by unified ItemsCombinator
- simplified Resolvers code (no need to work with variables/components/forms/filters independently)
- custom resolvers will be more resilient to future changes of template context (for example when we add another type of item to context)

Not sure about used names - if you have better ideas let me know.


lulco (approved):
I think I like it :) Have I rework my custom resolver now?

MartinMystikJonas:
Depends. If you are adding custom templates they yes. If you just overwrites some methods then it should not be necessary

Issue #198, opened 2023-01-05
Links with additional parameters


MartinMystikJonas:
When you create link you can add additional parameter that are not part of action/render method. For example you can reference some parsistent parameter but parameter can be also loaded dynamically in action body using `->getParameter()`. Should we ignore additional parameters?

lulco:
If we will ignore additional parameters, we can lost real errors or not?

MartinMystikJonas:
In some cases yes. At least we should ignore any parameters with matching persistent parameter in target presenter. That is definetly not an error. For additional parameters I think we maybe should emit specific error something like "Unknown parameter ABC for XX::YY" that can be ignored

lulco:
So we should collect persistent parameters, send them to LinkNodeProcessor and then ignore them if they are used in links as extra parameters?

can you prepare PR? I don’t use persistent parameters because I don’t like them from the first time is saw them :) so I really don’t know how it works.

lulco:
We can close this issue. If it was only partially solved, we will create new one.

MartinMystikJonas:
There is still issue that it is valid to  pass additional parameters ebcaise they can be read dynamically but it should have better error

lulco:
We can't cover 100% :) Focus on main issues from our (my, your and spaze's) projects :) If somebody else will report issue which is relevant, we will try to fix it. Or recommend to ignore it 🤷

lulco:
OK we failed to this issue :)

lulco:
`{link Do:something, 'filter' => ['a' => 'b']}`

maybe we could skip named parameters which are not parameters of target method

MartinMystikJonas:
Well either ignore it or show better error message that it is unknown parameter. When it is parameter not defined in method nor as persistent param there is no way to tell if it is typo in parameter or dynamic aprameter.

lulco:
Nette is adding those parameters to url query. All of them. Regardless they are used (read dynamically) or not

MartinMystikJonas:
so ignore it completely?

lulco:
If it is named and not matched with method param than yes

PR #235, opened 2023-01-12
Persistent params in links support

MartinMystikJonas:
Partialy solve #198

### DIFF
--- a/src/LinkProcessor/LinkParamsProcessor.php
+++ b/src/LinkProcessor/LinkParamsProcessor.php
@@ -12,6 +12,7 @@ use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\ArrayItem;
 use PhpParser\Node\Scalar\String_;
 use PHPStan\BetterReflection\BetterReflection;
+use ReflectionProperty;
 
 final class LinkParamsProcessor
 {
@@ -55,11 +56,13 @@ final class LinkParamsProcessor
             }
         }
 
-        $i = 0;
-        $reflectionMethod = (new BetterReflection())->reflector()->reflectClass($class)->getMethod($method);
+        $reflectionClass = (new BetterReflection())->reflector()->reflectClass($class);
+        $reflectionMethod = $reflectionClass->getMethod($method);
         if ($reflectionMethod === null) {
             throw new InvalidArgumentException("Method $class::$method not found");
         }
+
+        $i = 0;
         $methodParameters = [];
         foreach ($reflectionMethod->getParameters() as $param) {
             $name = $param->getName();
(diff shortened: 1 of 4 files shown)


src/LinkProcessor/LinkParamsProcessor.php
```diff
@@ -83,6 +86,16 @@ public function process(string $class, string $method, array $params): array
             }
         }
 
+        // remove persistent parameters
```
lulco, line 89: We should remove them only if they are not part of method params, I don't see the check here.

Let's say we have persistent param $p, and some actionDefault($a, $b, $p) <- here we should keep $p in url params

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Links/default.latte
```diff
@@ -85,3 +85,7 @@
 <a href="{link this#anchor}">Link to this with anchor</a>
 <a href="{link this, foo: bar, bar: baz}">Link to this with params</a>
 <a href="{link this#anchor, foo: bar, bar: baz}">Link to this with params and anchor</a>
+
+<a href="{link create persistent1: "persistent"}">With persistent annotation</a>
```
lulco, line 89: We need also create tests where persisten parameter is part of action e.g. `actionSomething($persistent)` <- here we want to keep persistent parameter.

src/LinkProcessor/LinkParamsProcessor.php
MartinMystikJonas: I think it is processed in previous block if they are part of method params and mobed to $methodParams

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Links/default.latte
MartinMystikJonas: Ok. Have to check how this conflict is handled by nette

MartinMystikJonas:
Idea how to best have dirrerent expexted errors based on Php version?

src/LinkProcessor/LinkParamsProcessor.php
lulco: Yes, but here you will remove them

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Links/default.latte
lulco: Do you think nette get persistent parameters in link as some special parameter?

src/LinkProcessor/LinkParamsProcessor.php
MartinMystikJonas: From transferedparams not from methodparams

lulco: yes, and then they are not returned to LinkProcessor

lulco:
> Idea how to best have dirrerent expexted errors based on Php version?

I think I need more context

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Links/default.latte
MartinMystikJonas: No. But i am not sure how request is provesed when method param and presistent param conflicts.

MartinMystikJonas:
> > Idea how to best have dirrerent expexted errors based on Php version?
> 
> I think I need more context

In PHP7.4 attributes do not exists so persistent paramater with attribute is seen as not-persistent. There is one more expected error. I fixed it by creating expected errors before abalyse call and conditionally add item there.

lulco:
If user analyses his php8.1 code with 7.4, he will get error 🤷 I would ignore this case

MartinMystikJonas:
Well I do not think it is something we should solve. If PHPStan runs on older version that app is imho something that should be fixed by user.

lulco (approved):

PR #238, opened 2023-01-12
Template cache key dependencies

MartinMystikJonas:
Fixes overwriting of compiled templates

Cache Key now include
- extension config
- PHP version
- Latte version
- instaled composer packages versions (if composer autolaoder is used)

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -69,7 +69,7 @@ services:
 
     - Efabrica\PHPStanLatte\Analyser\FileAnalyserFactory
     - Efabrica\PHPStanLatte\Analyser\AnalysedTemplatesRegistry(%analysedPaths%, %latte.reportUnanalysedTemplates%)
-    - Efabrica\PHPStanLatte\Compiler\LatteToPhpCompiler(%latte.tmpDir%)
+    - Efabrica\PHPStanLatte\Compiler\LatteToPhpCompiler(%latte.tmpDir%, ::md5(::json_encode(%latte%)))
 
     -
         factory: Efabrica\PHPStanLatte\Compiler\Postprocessor
--- a/src/Compiler/LatteToPhpCompiler.php
+++ b/src/Compiler/LatteToPhpCompiler.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace Efabrica\PHPStanLatte\Compiler;
 
+use Composer\InstalledVersions;
 use Efabrica\PHPStanLatte\Compiler\Compiler\CompilerInterface;
 use Efabrica\PHPStanLatte\Template\Template;
 use InvalidArgumentException;
+use Latte\Engine;
 
 final class LatteToPhpCompiler
 {
     private string $tmpDir;
+
+    private string $cacheKey;
 
     private CompilerInterface $compiler;
 
@@ -18,10 +22,16 @@ final class LatteToPhpCompiler
 
     public function __construct(
         ?string $tmpDir,
+        string $cacheKey,
         CompilerInterface $compiler,
         Postprocessor $postprocessor
     ) {
         $this->tmpDir = $tmpDir ?? sys_get_temp_dir() . '/phpstan-latte';
+        $this->cacheKey = $cacheKey . md5(
+            Engine::VERSION_ID .
+            PHP_VERSION_ID .
+            (class_exists(InstalledVersions::class) ? json_encode(InstalledVersions::getAllRawData()) : '')
+        );
         $this->compiler = $compiler;
         $this->postprocessor = $postprocessor;
     }
@@ -37,7 +47,12 @@ final class LatteToPhpCompiler
         $phpContent = $this->postprocessor->postProcess($phpContent, $template);
         $templateDir = pathinfo($templatePath, PATHINFO_DIRNAME);
         $templateFileName = pathinfo($templatePath, PATHINFO_BASENAME);
-        $contextHash = md5($template->getSignatureHash() . $context);
+        $contextHash = md5(
+            $templateContent .
+            $template->getSignatureHash() .
+            $context .
+            $this->cacheKey
+        );
 
         $replacedPath = getcwd() ?: '';
         if (strpos($templateDir, $replacedPath) === 0) {


src/Compiler/LatteToPhpCompiler.php
```diff
@@ -37,7 +43,14 @@ public function compileFile(Template $template, string $context = ''): string
...
         $templateDir = pathinfo($templatePath, PATHINFO_DIRNAME);
         $templateFileName = pathinfo($templatePath, PATHINFO_BASENAME);
-        $contextHash = md5($template->getSignatureHash() . $context);
+        $contextHash = md5(
+            $template->getSignatureHash() .
+            $context .
+            $this->cacheKey .
+            Engine::VERSION_ID .
+            PHP_VERSION_ID .
+            (class_exists(InstalledVersions::class) ? json_encode(InstalledVersions::getAllRawData()) : '')
```
lulco: Is this Installed Version from app or phpstan?

extension.neon
```diff
@@ -69,7 +69,7 @@ services:
 
     - Efabrica\PHPStanLatte\Analyser\FileAnalyserFactory
     - Efabrica\PHPStanLatte\Analyser\AnalysedTemplatesRegistry(%analysedPaths%, %latte.reportUnanalysedTemplates%)
-    - Efabrica\PHPStanLatte\Compiler\LatteToPhpCompiler(%latte.tmpDir%)
+    - Efabrica\PHPStanLatte\Compiler\LatteToPhpCompiler(%latte.tmpDir%, ::md5(::json_encode(%latte%)))
```
lulco, line 72: Do we need md5 here? In php code there is second one.

MartinMystikJonas: I pass cachekey as string and i do not want to pass big string when it is not necessary

src/Compiler/LatteToPhpCompiler.php
MartinMystikJonas: This is class defined by composer autoloader. It gives you info about what packages composer installed.

extension.neon
lulco: I don’t know @spaze is an expert here but I think I read something about double hashing as not good practise. Just not sure if it is more possible conflicts then

MartinMystikJonas: Double hashing is problem if you want to hide plaintext because counter intuitive it can make reverting it easier.

Bit for cahe it is imho ok. And also sexond hashong uses diffetent input

src/Compiler/LatteToPhpCompiler.php
lulco: I know, but I have phpstan installed as standalone app and not in tested application so if  I run it I’m not sure what packages will be loaded

extension.neon
lulco: 🤷

src/Compiler/LatteToPhpCompiler.php
MartinMystikJonas: Inam not sure too. But i would guess global version. Because it is what is used to load phpstan. But i think it is what we want because it is also what we use to compile.

lulco: Ah I thought you want to know nette packages' versions

lulco: I’ll try it which composer it will use

MartinMystikJonas: Basically I want to know if anythong that can affect how compiled version looks like changes.

lulco: Well it is affected by NodeVisitors of this extension and also by nette app itself - for example last change in nette/forms

MartinMystikJonas: So you use nette from app but PHPStan is global? How does autoloading work in such cases?

extension.neon
spaze: Doing md5(md5($foo)) wouldn't be ideal security-wise but in this case it's probably fine. Due to copy-on-write it may actually consume more memory than when passing just %latte% and doing the serialization when needed. (Update: scratch that, I guess the array will be passed as an array, not as a variable, so CoW would be irrelevant)

MartinMystikJonas: I need to remember passed value

src/Compiler/LatteToPhpCompiler.php
lulco: Yes. I think phpstan don’t need autoloading anymore. But before I just add paths to bootstrap.php to phpstan config.

MartinMystikJonas: Well it needs to autoload Latte classes somehow to use to compile templates.

lulco: they are included in this package https://github.com/efabrica-team/phpstan-latte/blob/main/composer.json#L9

MartinMystikJonas: So if you use global version you use latte version independent from your app to compile templates. This code woukd get versions of latte used to compile in global install - which is what we want.

lulco: I checked this now and I have both (phpstan global app and also nette app) there, so this is what I want, I think. Because not only latte/latte is important, but also nette/forms, nette/application and other packages which can register macros / extensions / providers.

Just maybe consider if we shouldn't do this only once, not for all templates compiled.

lulco (approved):

src/Compiler/LatteToPhpCompiler.php
```diff
@@ -4,24 +4,30 @@
...
     private Postprocessor $postprocessor;
 
     public function __construct(
         ?string $tmpDir,
+        string $cacheKey,
         CompilerInterface $compiler,
         Postprocessor $postprocessor
     ) {
         $this->tmpDir = $tmpDir ?? sys_get_temp_dir() . '/phpstan-latte';
+        $this->cacheKey = $cacheKey;
```
lulco: I would add it here = $cacheKey . (class_exists(InstalledVersions::class) ? json_encode(InstalledVersions::getAllRawData()) : '');

src/Compiler/LatteToPhpCompiler.php
MartinMystikJonas: Good point. We shoukd xompute cache key in constructor from static things and just combine it with template specific things in compile

PR #240, opened 2023-01-12
Adding a line about PHPStan levels and rules

### LINKED ISSUE #239: Add a line about PHPStan levels to docs
I'm running phpstan-latte with a different separate PHPStan config so I can test templates in parallel with running "regular" PHPStan (which runs much faster). And the config file has `customRulesetUsed: true` and no level specified which essentially means PHPStan is using level 0 - which means that possibly undefined variables are not reported.

I think it would be nice to add a line or two to docs saying that maybe you want to run PHPStan with rule level at least 1 :-) I'll now increase the level in my own config and see what I need to do and fix.

_Originally posted by @spaze in https://github.com/efabrica-team/phpstan-latte/issues/193#issuecomment-1380746804_

spaze:
Based on my brief (so far) usage I think this is a very high-level description of what the extension does. It should help users to better understand the functionality (as tested on me personally).

Close #239

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
-# PHPStan latte
-PHPStan extension to check compiled latte templates in context of Presenter or Component etc.
-It is based on Tomas Votruba's [blog series](https://tomasvotruba.com/blog/stamp-static-analysis-of-templates/) and his packages symplify and reveal.
+# PHPStan Latte extension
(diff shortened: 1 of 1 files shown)


lulco (approved):

lulco:
Thanks!

spaze:
Thank you too 😊

PR #247, opened 2023-01-13
Prepared forms “infrastracture”

lulco:


MartinMystikJonas:
I would avoid traits if possible.

Imho form shoukd extend container (as it do in nette) and ControlHolderBehaviour should be there. Or create common ancestor.

lulco:
> I would avoid traits if possible.

I want to deduplicate the code. For NameAndType interface I would also create trait. Its multiplicated throught whole project at least 5-6 times.

> Imho form shoukd extend container (as it do in nette) and ControlHolderBehaviour should be there. Or create common ancestor.

I've analysed it for long time and I decided to create ControlHolders. If Form would be child of Container, it will be also Control so it could be added to Form as Control and I believe its wrong so it shouldn’t be possible.

MartinMystikJonas:
How is it done in nette? I would mimic parent/interface relations as their are in nette here.

MartinMystikJonas:
I woukd go for this structure:
`abstract class AbstractControlHolder`
`Interface Control`
`class Container extends AbstractControlHolder implements Control`
`class Form extends AbstractControlHolder`
`class Field implements Control`

lulco:
What about groups? they are also ControlHolders, but the don't have type.

We don't have to copy Nette's structure. For us, container is kind of field / control which can have subfields. In Nette Container isn't Control.

MartinMystikJonas:
I just try to avoid traits whenever possible. But lets keep it this way for now. It is just matter of taste.

MartinMystikJonas (approved):

Issue #233, opened 2023-01-12
"Presenter::actionWithParametersMissingCorrespondingMethod not found"


spaze:
@spaze can you please paste compiled code for this template Admin/Presenters/templates/Talks/default.latte  ? 

Error should be different and transformed to more readable version.

```
 ------ ----------------------------------------------------------------------- 
  Line   Admin/Presenters/templates/Talks/default.latte rendered from           
         MichalSpacekCz\Admin\Presenters\InterviewsPresenter::interview         
         included in Admin/Presenters/templates/Interviews/default.latte        
         included in Admin/Presenters/templates/Interviews/interview.latte      
 ------ ----------------------------------------------------------------------- 
  -1     Method                                                                 
         MichalSpacekCz\Admin\Presenters\InterviewsPresenter::slidesWithParame  
         tersMissingCorrespondingMethod not found                               
 ------ ----------------------------------------------------------------------- 
```
([Link to the error in workflow run](https://github.com/spaze/michalspacek.cz/actions/runs/3897801662/jobs/6655848292#step:6:54))

_Originally posted by @lulco in https://github.com/efabrica-team/phpstan-latte/issues/193#issuecomment-1379648081_

spaze:
> @spaze can you please paste compiled code for this template Admin/Presenters/templates/Talks/default.latte ?
> 
> Error should be different and transformed to more readable version.

Yeah, that's indeed an interesting error :-) Here you go https://gist.github.com/spaze/ae214b8c1f1e5ba5816889d09e09cf80

lulco:
Sorry I ment compiled and transformed by our extension. It should be in /tmp folder if you didn’t change it. See https://github.com/efabrica-team/phpstan-latte/blob/main/tests/config.neon#L3

You can use tmp or temp as the value and files will be stored in your application for easier debugging

spaze:
Ah, sorry, https://gist.github.com/spaze/ae214b8c1f1e5ba5816889d09e09cf80 updated with hopefully the right files.

lulco:
@spaze see https://github.com/efabrica-team/phpstan-latte/pull/248, we think you should rework that template :) easy fix is to add Presenter to links

spaze:
Thanks, I have extracted the common part.

PR #248, opened 2023-01-13
Catching and transforming Invalid link error

lulco:
Related to https://github.com/efabrica-team/phpstan-latte/issues/233

It is not fixing the issue, just transforming an error to more readable version.

```
Invalid link: Unable to pass parameters to "MichalSpacekCz\Admin\Presenters\InterviewsPresenter::slides()", missing corresponding method.                                                  
💡 Add method actionSlides or renderSlides with corresponding parameters to presenter MichalSpacekCz\Admin\Presenters\InterviewsPresenter
```

@spaze @MartinMystikJonas 

Do anybody know how it works when in
Interview/default.latte we include Talks/default.latte and in it there is 
```
<a n:href="slides $item->talkId">Slides</a>
```
?

We consider that "actual presenter" is InterviewPresenter, but it looks like after include "actual presenter" is changed to TalksPresenter. I've never use it this way so I'm not sure what is correct.

"Fix" is to use full path in href:
```
<a n:href="Talks:slides $item->talkId">Slides</a>
```

### DIFF
--- a/src/Compiler/NodeVisitor/LinkNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/LinkNodeVisitor.php
@@ -19,6 +19,7 @@ use PhpParser\Node\Expr\PropertyFetch;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Stmt\Echo_;
 use PhpParser\NodeVisitorAbstract;
+use Throwable;
 
 /**
  * changed output from:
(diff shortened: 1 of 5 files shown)


MartinMystikJonas:
I think presenter stays the same.

MartinMystikJonas:
This is sort of valid error.

Interview.latte imports (not includes) talk.latte but then uses just one block from it. This error arises because different block contains link that would fail for InterviewPresenter. But that block is never called so error do not appear on web.

But it is still import of template that contains blocks invalid for current presenter.

I do not think there is a reasonable way we can eliminate this kind of errors. We would have to know what blocks can be used and with dynamic blocks it is impossible to tell.

Imho it should be fixed by extracting common block of interview and talk to separate template imported by both.

MartinMystikJonas (approved):

spaze:
What @MartinMystikJonas said.

PR #250, opened 2023-01-14
Use GitHub Actions checkout action v3 to resolve Node.js 12 deprecation warnings

spaze:
GitHub Actions workflow runs throw this:

> Composer outdated - PHP 8.0
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

This will stop it,

---
In other news:
image

### DIFF
--- a/.github/workflows/codesniffer.yml
+++ b/.github/workflows/codesniffer.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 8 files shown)


lulco:
Thanks!

PR #253, opened 2023-01-16
Caching compiled templates and line mapping refactoring

MartinMystikJonas:
Compiled teamplates ar cahce and not compiled every time when it is not needed (teste in my app and analysis time went from 1:45 to 1:20)

LineMap depended on compilation and was stored as global state. I reworked it as part of ErrorBuilder without global state and added caching too.


lulco:
I think the main problem is that we collect everything in compiled templates. That will improve speed more.

MartinMystikJonas:
That is also something I plan to fix

lulco (approved):

PR #254, opened 2023-01-16
Use only required collectors on compiled templates to collect includes

MartinMystikJonas:
I reuse LatteContextAnalyser with changed colelctors instead of creating new service. is it ok?

Do not seems like big change in tests. On ma computer difference is less than 1s on 40s run.


lulco:
well, new service would be much better, but... :) ok. 1s difference is not what I was expecting 🤔

lulco:
but it seems CI is much faster - 2m instead of 3m 40s

lulco (approved):

MartinMystikJonas:
Well maybe different bottleneck in CI than in my computer

PR #255, opened 2023-01-16
Better error messages for missing rendered/included templates

### LINKED ISSUE #252: Line -1 when included template not found
I have a template file with `{include "file that does not exist"}`, for example like this (notice the extra `}` after `.latte` and before `"`):

```html
<div class="column-sidebar">
	{include "../Post/common/sidebar.latte}"}
</div>
```

The error correctly says the file doesn't exist but the line shows `-1`:

```
 ------ ----------------------------------------------------------------------- 
  Line   Www/Presenters/templates/Talks/../Post/common/sidebar.latte} rendered  
         from MichalSpacekCz\Www\Presenters\TalksPresenter::talk included in    
         Www/Presenters/templates/Talks/talk.latte                              
 ------ ----------------------------------------------------------------------- 
  -1     Template file                                                          
         "/home/runner/work/michalspacek.cz/michalspacek.cz/site/app/Www/Prese  
         nters/templates/Talks/../Post/common/sidebar.latte}" doesn't exist.    
 ------ ----------------------------------------------------------------------- 
```

Can the line number be changed to where the wrong include is in the template?

MartinMystikJonas:
Resolves #252 
Depends on #253 but I ran rebase onto main if you want to wait with emrgin #253

### DIFF
--- a/src/Rule/LatteTemplatesRule.php
+++ b/src/Rule/LatteTemplatesRule.php
@@ -179,15 +179,23 @@ final class LatteTemplatesRule implements Rule
                     if ($includedTemplatePath[0] !== '/') {
                         $includedTemplatePath = $dir . '/' . $includedTemplatePath;
                     }
-                    $includedTemplatePath = realpath($includedTemplatePath) ?: $includedTemplatePath;
-                    $includeTemplate = new Template(
-                        $includedTemplatePath,
-                        $actualClass,
-                        $actualAction,
-                        $template->getTemplateContext()->mergeVariables($collectedTemplateRender->getVariables()),
-                        array_merge([$template->getPath()], $template->getParentTemplatePaths())
-                    );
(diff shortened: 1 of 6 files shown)


src/Rule/LatteTemplatesRule.php
```diff
@@ -200,6 +200,12 @@ private function analyseTemplates(array $templates, array &$errors, array &$alre
                         $templatePath,
                         $compileFilePath
                     );
+                } elseif (!is_file($$includedTemplatePath)) {
+                    $errors[] = $this->errorBuilder->buildError(
+                        new Error('Included latte template ' . $includedTemplatePath . ' does not exist.', $collectedTemplateRender->getFile(), $collectedTemplateRender->getLine()),
```
lulco: please add test for included path

lulco (approved):

PR #257, opened 2023-01-17
Add keywords to prompt users if they would like to add the package to require-dev

spaze:
If they run `composer require` without the `--dev` option, see https://getcomposer.org/doc/04-schema.md#keywords

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,9 +1,9 @@
 {
     "name": "efabrica/phpstan-latte",
     "license": ["MIT"],
-	"keywords": [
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,6 +1,9 @@
 {
     "name": "efabrica/phpstan-latte",
     "license": ["MIT"],
+	"keywords": [
```
lulco: please use spaces

lulco:
thanks!

PR #259, opened 2023-01-17
PHPStan extension installer support

MartinMystikJonas:


lulco:
Can you add info to readme like in https://github.com/phpstan/phpstan-nette?

```
If you also install [phpstan/extension-installer](https://github.com/phpstan/extension-installer) then you're all set!
Manual installation
```

MartinMystikJonas:
Docs changed based on phpstan-nette

README.md
```diff
@@ -21,23 +21,27 @@ The extension is based on Tomas Votruba's [blog series](https://tomasvotruba.com
...
 Add this line to your phpstan.neon:
 ```neon
 includes:
     - vendor/efabrica/phpstan-latte/rules.neon
 ```
 
+</details>
+
 It is recommended to use this extension with [phpstan-nette](https://github.com/phpstan/phpstan-nette) extension for better analysis of templates.
-Add also these lines to your phpstan.neon:
```
lulco, line 35: Do we want to remove this? User have to go to phpstan-nette and read its docs...

lulco (approved):

README.md
MartinMystikJonas: I think we should not repeat docs of another extensions.

PR #262, opened 2023-01-17
ContextResolver can create TemplateContext

MartinMystikJonas:


src/LatteContext/Resolver/ClassLatteContextResolver.php
```diff
@@ -38,6 +39,16 @@ public function getFilters(): array
         return $this->latteContext->filterFinder()->find($this->getClassName());
     }
 
+    protected function getTemplateContext(ReflectionClass $reflectionClass, LatteContext $latteContext): TemplateContext
```
lulco, line 42: where is this used?

lulco (approved):

src/LatteContext/Resolver/ClassLatteContextResolver.php
MartinMystikJonas: In our code nowhere. But it would be useful in custom resolvers (like mabar is writing).

lulco: shouldn't be public?

MartinMystikJonas: Oh shit it should

PR #267, opened 2023-01-18
LatteContextCollectorRegistry

MartinMystikJonas:


src/LatteContext/Collector/VariableCollector.php
```diff
@@ -64,7 +68,7 @@ public function collectData(Node $node, Scope $scope): ?array
         $isVariablesNode = false;
         $collectedVariables = [];
         foreach ($this->variableCollectors as $variableCollector) {
-            if (!$variableCollector->isSupported($node)) {
+            if (!in_array(get_class($node), $variableCollector->getNodeTypes())) {
```
lulco: This is not the same as instanceof

src/LatteContext/Collector/VariableCollector/SetParametersToTemplateVariableCollector.php
```diff
@@ -27,9 +27,9 @@ public function __construct(
         $this->templateTypeResolver = $templateTypeResolver;
     }
 
-    public function isSupported(Node $node): bool
```
lulco, line 30: Is supported was implemented more smart, it (can) check not just node types

MartinMystikJonas: At this point is was used just to check node types. We can put smarter code in collect method itself.

src/LatteContext/Collector/VariableCollector.php
MartinMystikJonas: True. it could be child type. So we should iterate over node types and use instanceof

src/LatteContext/Collector/VariableCollector/SetParametersToTemplateVariableCollector.php
MartinMystikJonas: Problem with smarter version is that we cannot optimize it. We have to call it for every single node of entire code even when it matches just handful assignments.

lulco: Well true, but as you said before. This wasn't big optimization anyway

MartinMystikJonas: It would be bigger on larger codebases I think. I would keep this version and we can add some version of isSupported as second level

lulco: ok let's try it

src/LatteContext/Collector/VariableCollector.php
lulco: this is fixed? if so, we can merge

lulco (approved):

src/LatteContext/Collector/TemplatePathCollector.php
```diff
@@ -63,7 +67,7 @@ public function collectData(Node $node, Scope $scope): ?array
         $paths = null;
         $isCollected = false;
         foreach ($this->templatePathCollectors as $templatePathCollector) {
-            if (!$templatePathCollector->isSupported($node)) {
+            if (!in_array(get_class($node), $templatePathCollector->getNodeTypes())) {
```
lulco: here should be also instance of

MartinMystikJonas:
It is fixed. I added abstraction for sub collectors and moved logic of chacking correct node type to isSupported. Childs can override isSupported if custom logic is needed.

lulco (approved):

PR #268, opened 2023-01-18
Some form labels are always true in if condition

lulco:
Failing test case https://github.com/efabrica-team/phpstan-latte/pull/268/commits/8b4b9e28472981c53e7d2cc60895d35e98ae36ac

POC for creating scope in NodeVisitor https://github.com/efabrica-team/phpstan-latte/pull/268/commits/0a7cefb89ea2c1ed5ecdacdc61a473413fd9c095

Comments in code

### DIFF
--- a/src/Compiler/NodeVisitor/CleanupNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/CleanupNodeVisitor.php
@@ -6,9 +6,11 @@ namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor;
 
 use Efabrica\PHPStanLatte\Resolver\NameResolver\NameResolver;
 use PhpParser\Node;
+use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\Closure;
 use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Stmt\Else_;
 use PhpParser\Node\Stmt\Expression;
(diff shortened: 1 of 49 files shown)


src/Compiler/NodeVisitor/CleanupNodeVisitor.php
```diff
@@ -6,28 +6,89 @@
...
+        $this->printerStandard = $printerStandard;
+        $this->nodeScopeResolver = clone $nodeScopeResolver;
+    }
+
+    public function beforeTraverse(array $nodes)
+    {
+        $content = $this->printerStandard->prettyPrintFile($nodes);
+        $filename = sys_get_temp_dir() . '/phpstan-latte-' . md5($content) . '.php';
+        file_put_contents($filename, $content);
+        require_once $filename;
```
lulco: we have to load this dummy file, because without it we can't analyse e.g. Form classes 
"Instantiated class Form_759c628d589650aaddca994fb1c153d3 not found."

But than we have to turn off require in LatteTemplateRule... Any idea?

src/Compiler/NodeVisitor/CleanupNodeVisitor.php
```diff
@@ -6,28 +6,89 @@
...
+    ) {
         $this->nameResolver = $nameResolver;
+        $this->scopeFactory = $scopeFactory;
+        $this->printerStandard = $printerStandard;
+        $this->nodeScopeResolver = clone $nodeScopeResolver;
+    }
+
+    public function beforeTraverse(array $nodes)
+    {
+        $content = $this->printerStandard->prettyPrintFile($nodes);
```
lulco: We can do this also in Postprocessor, and maybe we should. This is just POC.

src/Compiler/NodeVisitor/CleanupNodeVisitor.php
```diff
@@ -6,28 +6,89 @@
...
+        $filename = sys_get_temp_dir() . '/phpstan-latte-' . md5($content) . '.php';
+        file_put_contents($filename, $content);
+        require_once $filename;
+        $scope = $this->scopeFactory->create(ScopeContext::create($filename));
+
+        $nodeCallback = function (Node $node, Scope $scope): void {
```
lulco: callback can be created in each NodeVisitor (which will implement some ScopeInterface) and called somehow from Postprocessor

src/Compiler/NodeVisitor/CleanupNodeVisitor.php
```diff
@@ -6,28 +6,89 @@
...
 
     public function leaveNode(Node $node)
     {
+        if ($node instanceof If_) {
+            if ($node->cond instanceof Assign) {
+                if ($node->cond->expr instanceof MethodCall) {
+                    if ($this->nameResolver->resolve($node->cond->expr) !== 'getLabel') {
+                        return null;
+                    }
+                    $type = $node->cond->expr->getAttribute(self::METHOD_CALL_CALLER_TYPE);
```
lulco: Here we just use attribute created in nodeCallback

src/Compiler/NodeVisitor/CleanupNodeVisitor.php
```diff
@@ -6,28 +6,89 @@
...
+        $this->nodeScopeResolver->processNodes($nodes, $scope, $nodeCallback);
+
+        return null;
     }
 
     public function leaveNode(Node $node)
     {
+        if ($node instanceof If_) {
+            if ($node->cond instanceof Assign) {
+                if ($node->cond->expr instanceof MethodCall) {
```
lulco: this if is always true for form fields of type CheckboxList (and some other - I have to check return types)
```php
if ($label = $form['checkboxlist']->getLabel()) {
    echo $label;
}
```

src/Compiler/NodeVisitor/CleanupNodeVisitor.php
```diff
@@ -6,28 +6,89 @@
...
+                    if ($this->nameResolver->resolve($node->cond->expr) !== 'getLabel') {
+                        return null;
+                    }
+                    $type = $node->cond->expr->getAttribute(self::METHOD_CALL_CALLER_TYPE);
+                    if (!$type instanceof ObjectType) {
+                        return null;
+                    }
+
+                    if ($type->isInstanceOf('Nette\Forms\Controls\CheckboxList')->yes()) {
+                        return array_merge([new Expression($node->cond)], $node->stmts);
```
lulco: we just use condition and statements, so result will be
```php
$label = $form['checkboxlist']->getLabel()
echo $label;
```

MartinMystikJonas:
I will chack it in detail later but as general approach it could work by use of attributes. But I am not sure about performance impact here. NodeScopeResolver do basically whole analysis again.

I am thinking about different approach to ignore these errors. Maybe create custom rule that would detect this kind of errors in compiled templates and emits error "IGNORE: pattern" and then use ErrorBuilder to  cancel out original error based on pattern on given line.

Simply detect same error PHPStan is detecting, check if it is called on checkbox and if it is add error "IGNORE: Condition is always true" on same location.

It would not need another NodeScopeResolver run on compiled template but it will be just another rule used in normal run.

lulco:
> I will chack it in detail later but as general approach it could work by use of attributes. But I am not sure about performance impact here. NodeScopeResolver do basically whole analysis again.
> 
> I am thinking about different approach to ignore these errors. Maybe create custom rule that would detect this kind of errors in compiled templates and emits error "IGNORE: pattern" and then use ErrorBuilder to cancel out original error based on pattern on given line.
> 
> Simply detect same error PHPStan is detecting, check if it is called on checkbox and if it is add error "IGNORE: Condition is always true" on same location.
> 
> It would not need another NodeScopeResolver run on compiled template but it will be just another rule used in normal run.

I was thinking about it, but there is no such information about what is called by which class, it is just error "Condition is always true". We can't ignore all errors of this type. Or do you have some suggestion?

MartinMystikJonas:
My idea is that we create PHPStan rule that would have this info available. And that re would add "inverse error". So PHPstan rule would emit error X on line y. Our rule would emit "IGNORE: X" on line y. Ans error builder then would check all "IGNORE" error and remove matching normal errors

lulco:
I see, that's pretty complicated :) In ErrorBuilder there is only actual error, not all errors. And I wanted to use built-in phpstan rules, not create own rules for each thing.

Btw this errors seems to be fixed in latte 3, so maybe I'll just ignore it https://github.com/efabrica-team/phpstan-latte/actions/runs/3953621023/jobs/6770094462

MartinMystikJonas:
To summarise:

IgnoreLabelAlwaysTrueErrorRule - PHPStan rule that would detect always true condition on Label in compiled template (same check as in phpstan rule + it is on label + it is in class that is child of Template). If it detects this it will add error "IGNORE: If condition is always true" at that exact spot.

ErrorBuilder - iterate over all error. If "IGNORE: something is found" remove all (one?) errors "something is found" on given line.

So comopiled code would stay the same. PHPStan will edd normal error. But our rule would add error that would cancel out with this normal error so it would be ignored.

MartinMystikJonas:
It would use build in phpstan rules. We will just create new "ignore rules" for error we want to ignore.

Well if it is latte2 specific maybe just ignore it

lulco:
> 



> To summarise:
> 
> IgnoreLabelAlwaysTrueErrorRule - PHPStan rule that would detect always true condition on Label in compiled template (same check as in phpstan rule + it is on label + it is in class that is child of Template). If it detects this it will add error "IGNORE: If condition is always true" at that exact spot.
> 
> ErrorBuilder - iterate over all error. If "IGNORE: something is found" remove all (one?) errors "something is found" on given line.
> 
> So comopiled code would stay the same. PHPStan will edd normal error. But our rule would add error that would cancel out with this normal error so it would be ignored.

I know what did you tell the first time. I just think this is pointlessly complicated and we would have to create custom rule for all such errors and it is some second way how to solve errors in compiled template. I like more to change compiled template, but sometimes scope is needed 🤷 

I'm closing it for now and ignore this error locally.

lulco:
I'll rewrite it without using scope, will use just known parts of code - method name getLabel, assign to variable $ʟ_label. It is specific enough for this purpose I think

MartinMystikJonas:
Seems ok to me but tests fails

lulco:
Yes some prefer lowest is different, I’ll check

lulco:
There was different name of variable $_label in older versions. Fixed. I'm going to merge

PR #270, opened 2023-01-19
Postprocessor optimization

MartinMystikJonas:
Postprocessor did repeated transformations from code-to-nodes-to-code for every group of visitors. Changed to resuse of nodes and it seems it works and is faster (about 5-10% faster in my machine on tests). Was there any reason to go back to code or is this ok?


lulco:
The reason was that number of lines were changed when some group of node visitors remove / add nodes, start lines were changed. I had to save and load nodes again and re-parse content to nodes. I have no idea how it is possible it works now :) with this approach there are no groups needed or they are?

MartinMystikJonas:
You mean lines for mapping php line to latte line? It was reworked as part of compiled files cache so it seems it is not needed anymore.

I do not know what purpose groups have but I would fear that some visitors may conflict when tey are run all at once if they modify same paets. I would keep groups for now.

lulco (approved):
OK let's merge

PR #276, opened 2023-01-20
Assign to dynamic template variable

### LINKED ISSUE #201: Assign to dynamic template variable
$var = 'something';
$this->template->$var = 'value';

Should be evaluated as something variable sent to template.

this should work also if it is called in foreach

MartinMystikJonas:
Resolves #201


lulco:
what about
```
[$this->template->$var,] = ['foo', 'bar'];
```
or
```
$this->template->setParameters([
$var => 'foo',
]);
```
or
```
$this->template->add($var, 'foo');
```

Mechanism is probably the same, we just need to add it to all subcollectors

MartinMystikJonas:
Second and third should work already I think because we are resolving key names anyways and value resolver supports constant string as weall as variables with constant values. 

First one I shoudl check but i probably does not work now.

lulco:
create tests please and check all use cases

MartinMystikJonas:
Already working on it ;-)

MartinMystikJonas:
Done. It should work for all cases now.

src/LatteContext/Collector/VariableCollector/SetParametersToTemplateVariableCollector.php
```diff
@@ -55,6 +56,7 @@ public function collect(Node $node, Scope $scope): ?array
 
         $types = $scope->getType($node->args[0]->value);
         if (!$types instanceof ConstantArrayType) {
+            print_r($types->describe(VerbosityLevel::precise()));
```
lulco: forgotten print

lulco (approved):

PR #279, opened 2023-01-20
Added support for union types for components

lulco:


src/LatteContext/Collector/ComponentCollector.php
```diff
@@ -164,14 +165,31 @@ private function findAssignToThis(Assign $node, Scope $scope, ClassReflection $c
...
-        $components = [];
-
-        // TODO support union types
+        $componentType = null;
         if ($componentArgType instanceof ObjectType && $componentArgType->isInstanceOf('Nette\ComponentModel\IComponent')->yes()) {
+            $componentType = $componentArgType;
+        } elseif ($componentArgType instanceof UnionType) {
+            $componentTypes = [];
+            foreach ($componentArgType->getTypes() as $type) {
+                if ($type instanceof ObjectType && $type->isInstanceOf('Nette\ComponentModel\IComponent')->yes()) {
```
lulco, line 182: not sure if all types from union have to by IComponent or some of them :)

src/LatteContext/Collector/ComponentCollector.php
```diff
@@ -181,12 +199,7 @@ private function buildComponents(Node $node, Scope $scope, ClassReflection $clas
             }
         }
 
-        $lattePhpDoc = $this->lattePhpDocResolver->resolveForNode($node, $scope);
```
lulco, line 184: moved to method start

src/LatteContext/Collector/ComponentCollector.php
```diff
@@ -181,12 +199,7 @@ private function buildComponents(Node $node, Scope $scope, ClassReflection $clas
...
-        if ($lattePhpDoc->isIgnored()) {
-            return null;
-        }
         if ($lattePhpDoc->hasComponents()) {
-            $components = [];
```
lulco, line 189: other thing I'm not sure with... If I collect some components, and there are some in phpdoc I think I don't want to reset all collected. Just overwrite those which are in php doc and add missing from php doc

MartinMystikJonas: Correct we did same for variables

src/LatteContext/Collector/ComponentCollector.php
MartinMystikJonas: I would take what is component and ignore others

lulco: ok, than it is implemented this way

MartinMystikJonas (approved):

PR #281, opened 2023-01-20
Lower memory usage

MartinMystikJonas:


MartinMystikJonas:
Lowered parser cache. It should not be needed anyway

lulco (approved):
🤞

PR #286, opened 2023-01-21
Improved detection of terminating method calls

### LINKED ISSUE #119: Differentiate between sometimes and always called methods

### LINKED ISSUE #261: Need to have `redirectPermanent` in `earlyTerminatingMethodCalls`
I'm using phpstan/phpstan-nette 1.2.1, nette/application 3.1.10, phpstan-latte 0.4.0.

My action uses `redirectPermanent()`:

```php
	public function actionStoragesQuestions(): never
	{
		$this->redirectPermanent('PasswordsStorages:questions');
	}

```

The phpstan-latte extensions says
```
 ------ ------------------------------------------------------------ 
  Line   Pulse/Presenters/PasswordsPresenter.php                     
 ------ ------------------------------------------------------------ 
  22     Cannot resolve latte template for action storagesQuestions  
  33     Cannot resolve latte template for action storagesRating     
 ------ ------------------------------------------------------------ 
```

even though `redirectPermanent` has `return never`:
```php
	/**
	 * Permanently redirects to presenter, action or signal.
	 * @param  string   $destination in format "[//] [[[module:]presenter:]action | signal! | this] [#fragment]"
	 * @param  array|mixed  $args
	 * @return never
	 * @throws Nette\Application\AbortException
	 */
	public function redirectPermanent(string $destination, $args = []): void
	{
```

To get rid of those "Cannot resolve latte template" errors, I need to add this to my `phpstan.neon` config:

```neon
earlyTerminatingMethodCalls:
		Nette\Application\UI\Component:
			- redirectPermanent
```

It's not needed for "normal" PHPStan anymore (starting with nette/application 3.1.10) but phpstan-latte still requires it.

Can the extension check the return type hint too?

MartinMystikJonas:
Resolves #119 
Resolves #261

### DIFF
--- a/src/LatteContext/Finder/MethodCallFinder.php
+++ b/src/LatteContext/Finder/MethodCallFinder.php
@@ -92,7 +92,17 @@ final class MethodCallFinder
      */
     public function traverseCalled(callable $callback, string $className, string $methodName, string $currentClassName = null): array
     {
-        return $this->traverseInMethodCalls($callback, $className, $methodName, $currentClassName);
+        return $this->traverseInMethodCalls($callback, $className, $methodName, $currentClassName, false);
+    }
+
+    /**
+     * @template T
+     * @param callable(class-string, string, array<T[]>): T[] $callback
+     * @return T[]
+     */
+    public function traverseAlwaysCalled(callable $callback, string $className, string $methodName, string $currentClassName = null): array
+    {
+        return $this->traverseInMethodCalls($callback, $className, $methodName, $currentClassName, true);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


lulco (approved):
Isn't there some built-in method in phpstan itself?

MartinMystikJonas:
Didnt found any

lulco:
Something using parameter earlyTerminatingMethodCalls?

spaze:
> earlyTerminatingMethodCalls

That would be one way how to detect such methods but PHPStan uses `@return never` tags etc, too, and they seem to be preferred over populating the config param.

PR #287, opened 2023-01-22
CollectedPaths docs and by default collect everything

### LINKED ISSUE #256: CollectedPaths
- [ ] By default collect everywhere (maybe except nette and latte where it is pointless?)
- [ ] Add documentation

MartinMystikJonas:
Resolves #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ It is recommended to use this extension with [phpstan-nette](https://github.com/
 * [Link checking](docs/configuration.md#link-checking)
 * [Other configuration options](docs/configuration.md#other-configuration-options)
 
+If analysis reports errors about missing context (variables/components, ...) set in external packages you need to set option `latte.collectedPaths` as descibed in [Analyser configuration](docs/configuration.md#analyser-configuration).
+
 ## Annotations
 
 There are cases that cannot be resolved automatically by static analysis. 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -101,6 +101,19 @@ parameters:
 
 ## Analyser configuration
 
+### collectedPaths
+Type: `array`
+
+List of paths outside of analysedPaths that are scanned when latte context (variables, components, ...) is collected for template analysis.
+
+Example:
+```neon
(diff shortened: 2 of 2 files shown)


lulco:
I changed my mind here. I want this package to be as fast as possible, so we will analyse only analysedPaths by default. Of course we need documentation how to collect data from additional directories, but that's all. If user gets error about non-existing variable / component he will go throught docs and see how to set another dirs to collected paths, but if he will get no error but analysis will be slow, he will just nervous.

README.md
```diff
@@ -51,6 +51,8 @@ It is recommended to use this extension with [phpstan-nette](https://github.com/
 * [Link checking](docs/configuration.md#link-checking)
 * [Other configuration options](docs/configuration.md#other-configuration-options)
 
+If analysis is slow you should set option `latte.collectedPaths` as desciption in [Analyser configuration](docs/configuration.md#analyser-configuration).
```
lulco: - if you need to collect (not analyse - no errors will be reported) extra files or directories 
- typo description

docs/configuration.md
```diff
@@ -101,6 +101,23 @@ parameters:
 
 ## Analyser configuration
 
+### collectedPaths
+Type: `array`
+
+List of paths outside of analysedPaths that are analysed when latte context (variables, components, ...) is collected for template analysis.
```
lulco: They are not analysed in phpstan mean. We just collect data from them

docs/configuration.md
```diff
@@ -101,6 +101,23 @@ parameters:
...
+
+By default all files referenced from analysed code are collected. That can be slow if you have some big vendor directory.
```
lulco: Remove this part

MartinMystikJonas:
Ok so keep original implementation just update docs?

lulco:
> Ok so keep original implementation just update docs?

Yes please.

lulco (approved):

MartinMystikJonas:
Changed to jsut docs update

PR #289, opened 2023-01-23
Mark some errors as warnings

### LINKED ISSUE #285: Mark some errors as warnings and do not count them
We could implement it in ErrorFormatter. We will print those warnings but not count them.

- define patterns for warnings and use them in error formatter

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/285

Output will look like this:
image

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -148,7 +148,7 @@ parameters:
         reportUnanalysedTemplates: true
 ```
 
-## Ignoring errors
+## Errors
 
 ### errorPatternsToIgnore
 Type: `array`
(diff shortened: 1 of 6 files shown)


MartinMystikJonas:
Not sure what is added walue here,  I use our custom error formatter anyway :-) But it seems ok.

MartinMystikJonas (approved):

lulco:
added value is, that I will not ignore some errors completely. I will see them in warnings section, but build will be green :)

MartinMystikJonas:
Well I guess it is that my mindset is locked "no warnings allowed" since college :-)

PR #290, opened 2023-01-23
Reworked variables collectors

### LINKED ISSUE #163: Refactoring ideas

MartinMystikJonas:
Resolves #163


extension.neon
```diff
@@ -116,10 +116,10 @@ services:
             - addNodeVisitor(50, Efabrica\PHPStanLatte\Compiler\NodeVisitor\AddParametersForBlockNodeVisitor())
             - addNodeVisitor(50, Efabrica\PHPStanLatte\Compiler\NodeVisitor\CopyDefinedVarsToOtherMethodsNodeVisitor())
             - addNodeVisitor(100, Efabrica\PHPStanLatte\Compiler\NodeVisitor\AddTypeToComponentNodeVisitor())
-            - addNodeVisitor(100, Efabrica\PHPStanLatte\Compiler\NodeVisitor\AddVarTypesNodeVisitor())
+            - addNodeVisitor(100, Efabrica\PHPStanLatte\Compiler\NodeVisitor\AddVarTypesNodeVisitor(%latte.globalVariables%))
             - addNodeVisitor(100, Efabrica\PHPStanLatte\Compiler\NodeVisitor\RemoveExtractParamsNodeVisitor())
             - addNodeVisitor(100, Efabrica\PHPStanLatte\Compiler\NodeVisitor\RenderBlockNodeVisitor())
-            - addNodeVisitor(200, Efabrica\PHPStanLatte\Compiler\NodeVisitor\ChangeFiltersNodeVisitor())
+            - addNodeVisitor(200, Efabrica\PHPStanLatte\Compiler\NodeVisitor\ChangeFiltersNodeVisitor(...))
```
lulco, line 122: What does `…` mean here?

lulco (approved):
I think this PR could be omitted and focus on refactoring with https://github.com/efabrica-team/phpstan-latte/issues/143 where each Template has its own engine and it means there are no global variables, no filters, no macros etc. Everything depends on which engine is used. Also each nette template class (DefaultTemplate as default) has its own variables.
But if you already implement it… it's up to you if you will merge it. Just saying that this code will be reworked again sooner or later.

MartinMystikJonas:
I will merge it. When we willngo for customisable engines something similar to this would be needed anyway. Only some vars that are hardcoded now would be loaded based on engine/template in vartypesnodevisitor.

extension.neon
MartinMystikJonas: Autowire all constructor params

lulco: That's default, isn't it?

MartinMystikJonas: Seems like it do not work implicitly inside setups. I had to add it to prevent error

lulco: Interesting. We have 2 params in https://github.com/efabrica-team/phpstan-latte/blob/main/src/Compiler/NodeVisitor/LinkNodeVisitor.php and it works from setup

MartinMystikJonas: 🤔🤷

PR #292, opened 2023-01-24
exit() and die() evaluated as early terminating call

### LINKED ISSUE #291: exit(); is not evaluated as early terminating call
I have Presenter with action like:

```php
public function actionCopy(string $id): void
{
    // do something
    print $id;
    exit();
}
```

There is no template for this action obviously, but I've got error `Cannot resolve latte template for action copy`

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/291

### DIFF
--- a/src/LatteContext/Collector/MethodCallCollector.php
+++ b/src/LatteContext/Collector/MethodCallCollector.php
@@ -43,7 +43,7 @@ final class MethodCallCollector extends AbstractLatteContextCollector
     {
         return [
             CallLike::class,
-            Exit_::class
+            Exit_::class,
         ];
     }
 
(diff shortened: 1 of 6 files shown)


lulco:
@MartinMystikJonas I think phpstan count exit and die as early terminating calls

MartinMystikJonas:
We have to check different type of node

MartinMystikJonas:
Fixed but tests needs to be fixed too (changed line numbers and *Sometimes versions should be in expected collected)

lulco:
> Fixed but tests needs to be fixed too (changed line numbers and *Sometimes versions should be in expected collected)

I think I fixed it meantime. There is only sniffer failing now.

lulco:
Thanks for quick fix!

PR #299, opened 2023-01-26
Rewritten TerminatingCallResolver to use phpstan’s built-in method

lulco:
I found this method in NodeScopeResolver and I used it. Tests are green, so I think we should keep it. Can you please check other places where early terminating is collected and use this method instead of some custom written?

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -107,7 +107,7 @@ services:
 
     # Method resolvers
     - Efabrica\PHPStanLatte\Resolver\CallResolver\CalledClassResolver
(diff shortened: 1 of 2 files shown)


MartinMystikJonas (approved):
Not sure how it detects enevr type but if it works lets keep it

lulco:
Maybe this?
```php
        $exprType = $scope->getType($expr);
        if ($exprType instanceof NeverType && $exprType->isExplicit()) {
            return true;
        }
```

PR #301, opened 2023-01-27
Template resolver helpers

MartinMystikJonas:


src/LatteContext/LatteContext.php
```diff
@@ -115,6 +115,16 @@ public function templateRenderFinder(): TemplateRenderFinder
         return $this->templateRenderFinder;
     }
 
+    public function getClassTemplateContext(string $className): TemplateContext
+    {
+        return new TemplateContext(
+            $this->variableFinder()->find($className),
```
lulco, line 121: this returns only "global" class variables, components, forms and filters right?

lulco: can ClassTemplate have form, component, filter? I don't think so

MartinMystikJonas: Components/forms from createComponent are registered as global class.

Everything can be global class from annotation

lulco: I ment something else. If mabar has TemplateClasses and pass them to this method he will not get forms, components and filters, only variables.

lulco (approved):

src/LatteContext/LatteContext.php
MartinMystikJonas: He going to extract control class from template class and then he will pass control class here. To extract variables from template class LatzeContextHelper is used.

lulco (approved):

PR #303, opened 2023-01-30
Resolve generics template types in latte context

MartinMystikJonas:
Fixed bug when variable/component/... uses template type from generics

### DIFF
--- a/src/Compiler/TypeToPhpDoc.php
+++ b/src/Compiler/TypeToPhpDoc.php
@@ -6,8 +6,10 @@ namespace Efabrica\PHPStanLatte\Compiler;
 
 use PHPStan\PhpDoc\TypeStringResolver;
 use PHPStan\Type\ErrorType;
+use PHPStan\Type\Generic\TemplateType;
 use PHPStan\Type\StaticType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\VerbosityLevel;
 use Throwable;
 
@@ -25,6 +27,14 @@ final class TypeToPhpDoc
         if ($type instanceof StaticType) {
             $type = $type->getStaticObjectType();
         }
+
+        // replace unresolved template types with their bounds (T of stdClass -> stdClass)
+        $type = TypeTraverser::map($type, static function (Type $type, callable $traverse): Type {
+            if ($type instanceof TemplateType) {
+                return $traverse($type->getBound());
+            }
+            return $traverse($type);
+        });
 
         $phpDoc = $type->describe(VerbosityLevel::precise());
         try {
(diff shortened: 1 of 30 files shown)


src/Template/ItemCombinator.php
```diff
@@ -35,16 +36,26 @@ public static function union(array ...$itemArrays): array
...
             }
         }
         return array_values($union);
     }
+
+    /**
+     * @param NameTypeItem[] $items
+     * @return NameTypeItem[]
+     */
+    public static function resolveTemplateTypes(array $items, string $declaringClass, ?string $currentClass): array
```
lulco, line 52: how is template type resolver connected with item combinator?

src/Compiler/TypeToPhpDoc.php
```diff
@@ -26,6 +28,14 @@ public function toPhpDocString(Type $type): string
             $type = $type->getStaticObjectType();
         }
 
+        // replace unresolved template types with their bounds (T of \StdClass -> \StdClass)
```
lulco: stdClass and no slash

src/LatteContext/CollectedData/CollectedMethodCall.php
```diff
@@ -15,11 +15,14 @@ final class CollectedMethodCall extends CollectedLatteContextObject
     public const TERMINATING_CALL = 'terminating';
     public const OUTPUT_CALL = 'output';
 
-    private string $callerClassName;
+    /**
+     * @var ?class-string
+     */
+    private ?string $callerClassName;
```
lulco, line 21: why nullable?

MartinMystikJonas: Theoretically it can be called outside class scope. We used '' for it originally but nullable is imho better.

lulco: I see it now, ok null it is

src/Template/ItemCombinator.php
MartinMystikJonas: Well only vaguely bit I did not find better place for it. I think af it as sort of combining Items with current template. Or would you preffer separate ItemTemplateTypeResolver? Or something like ItemHelper?

lulco: What is this for? I don't get it. You use it in component finder, filter finder etc.

MartinMystikJonas: When we collect types we might collect `T` reffering to generic template type `@template T of Something`. We then need to replace `T` with actual type passed to generic in given implementation of generic type.

MartinMystikJonas: We cannor do it at collect time because when we are collecting types from generic parent it can be used in many different childs. We have to replave it in finder when we know actualClass

lulco: what is array $items for? I get lost because of them :D

MartinMystikJonas: It is for example array or Variable and this method check types of all variables in array and resolves genetic template types with actual types. So tresult is array of Variable with their types resolved.

lulco (approved):

PR #304, opened 2023-01-30
Presenter factory bootstrap

### LINKED ISSUE #27: Automatic detection of application mapping
Maybe add option applicationConfig to specify config file to load mapping config from? Just parse it as .neon and load application.mapping key?

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/27

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -214,7 +214,7 @@ Example:
 ```neon
 parameters:
     latte:
-        engineBootstrap: presenter.factory.php
+        presenterFactoryBootstrap: presenter.factory.php
 ```
 
 ### applicationMapping
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -205,6 +205,18 @@ parameters:
 
 ## Link checking
 
+### presenterFactoryBootstrap
+Type: `string`
+
+If provided return value of this php file is used as Presenter Factory.
+
+Example:
+```neon
+parameters:
+    latte:
+        engineBootstrap: presenter.factory.php
+```
+
 ### applicationMapping
 Type: `array`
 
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -205,6 +205,18 @@ parameters:
...
+### presenterFactoryBootstrap
+Type: `string`
+
+If provided return value of this php file is used as Presenter Factory.
+
+Example:
+```neon
+parameters:
+    latte:
+        engineBootstrap: presenter.factory.php
```
MartinMystikJonas: Wrong parameter

src/LinkProcessor/PresenterFactoryFaker.php
```diff
@@ -4,25 +4,42 @@
...
+        $this->presenterFactoryBootstrap = $presenterFactoryBootstrap;
     }
 
-    public function getPresenterFactory(): PresenterFactory
+    public function getPresenterFactory(): ?IPresenterFactory
     {
-        $presenterFactory = new PresenterFactory();
-        $presenterFactory->setMapping($this->mapping);
-        return $presenterFactory;
+        $presenterFactory = null;
```
MartinMystikJonas, line 34: If presenter factory is already created it should be returned amd not created again

lulco: Yeah, I have presenterFactory property, but only initialized, not used 🤦

docs/configuration.md
lulco: thanks

MartinMystikJonas (approved):

spaze:
Can you add some more documentation please? Like when would I want to use it? And if I do, do I still need to use `applicationMapping`? Also an example of what should the file look like would help, like the example file contents in `engineBootstrap` description. Thanks!

lulco:
Updated in https://github.com/efabrica-team/phpstan-latte/pull/307

PR #306, opened 2023-01-30
Filters with FilterInfo

### LINKED ISSUE #296: Filters with FilterInfo
Latte filters can be defined with FilterInfo as first parameter (it is automatically added by FilterExecutor)

```php
public function markdown(FilterInfo $info, $text) {
```

But when we replace filtrs with methods we do not add it. When called filter first argument is of type FilterInfo we should pass it.

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/296

### DIFF
--- a/src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
@@ -10,6 +10,7 @@ use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\FiltersNodeVisitorBehavi
 use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\FiltersNodeVisitorInterface;
 use Efabrica\PHPStanLatte\Compiler\TypeToPhpDoc;
 use Efabrica\PHPStanLatte\Template\Variable;
+use Latte\Runtime\FilterInfo;
 use PhpParser\Comment\Doc;
 use PhpParser\Node;
 use PhpParser\Node\Arg;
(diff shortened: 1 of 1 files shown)


src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
```diff
@@ -238,4 +241,20 @@ private function createFilterCallNode(string $filterName, array $args): ?Node
...
+    /**
+     * @param ReflectionFunction|ReflectionMethod $reflection
+     * @param Arg[]|VariadicPlaceholder[] $args
+     * @return Arg[]|VariadicPlaceholder[]
+     */
+    private function updateArgs($reflection, array $args): array
+    {
+        $parameter = $reflection->getParameters()[0] ?? null;
+        if ($parameter instanceof ReflectionParameter && $parameter->getType() instanceof ReflectionNamedType && $parameter->getType()->getName() === FilterInfo::class) {
+            $args = array_merge([
```
MartinMystikJonas, line 254: array_unshift would be cleaner

src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
```diff
@@ -28,6 +29,10 @@
 use PhpParser\Node\VariadicPlaceholder;
 use PhpParser\NodeVisitorAbstract;
 use PHPStan\BetterReflection\BetterReflection;
+use PHPStan\BetterReflection\Reflection\ReflectionFunction;
```
MartinMystikJonas, line 32: We try to replace direct use of BetterREflection with PHPStan ReflectionProvider when possible for performance reasons

lulco: In this file, BetterReflection was used so I continue with it. I don’t want to mix things in pull request. For replacing there is another issue, when it can be done as the only thing at that PR.

src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
lulco: This part is unchanged, for me array_merge is more clear. But I can change it to array_unshift

MartinMystikJonas: If you like merge more then keep it

src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
MartinMystikJonas: if it is already used then keep it we will repalce it later

MartinMystikJonas (approved):

PR #307, opened 2023-01-30
Updated docs for presenter factory bootstrap

lulco:
@spaze can you please check this if it is understandable?


spaze:
Yeah, it's clear now, thank you!

PR #310, opened 2023-02-03
Multipler components 

MartinMystikJonas:
Some componenty can be created dynamically using Multiplier.

I wrote basic support but there is problem how to determine what type is component created by Multiplier.

We proably would need to make Multiplier generic and use tempalte type to specify what type created components are. But this should be supported in nette or phpstan-nette I think

### DIFF
--- a/src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
+++ b/src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior;
 
 use Efabrica\PHPStanLatte\Template\Component;
+use PHPStan\Type\ObjectType;
 
 trait ComponentsNodeVisitorBehavior
 {
@@ -32,6 +33,11 @@ trait ComponentsNodeVisitorBehavior
             }
             if (count($componentNameParts) === 0) {
                 return $component;
+            }
+            $componentType = $component->getType();
+            if ($componentType instanceof ObjectType && $componentType->isInstanceOf('Nette\Application\UI\Multiplier')->yes()) {
+                // TODO: How to determine component type?
+                return new Component('', new ObjectType('Nette\Application\UI\Control'));
             }
             return $this->findComponentByName($component->getSubcomponents() ?: [], implode('-', $componentNameParts));
         }
--- a/src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
+++ b/src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
@@ -5,7 +5,13 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior;
 
 use Efabrica\PHPStanLatte\Template\Component;
-use PHPStan\Type\ObjectType;
+use PhpParser\ConstExprEvaluationException;
+use PhpParser\ConstExprEvaluator;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\ConstFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Scalar\Encapsed;
+use PhpParser\Node\Scalar\EncapsedStringPart;
 
 trait ComponentsNodeVisitorBehavior
 {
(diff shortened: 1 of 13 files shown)


src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
```diff
@@ -33,6 +34,11 @@ private function findComponentByName(array $components, string $componentName):
             if (count($componentNameParts) === 0) {
                 return $component;
             }
+            $componentType = $component->getType();
+            if ($componentType instanceof ObjectType && $componentType->isInstanceOf('Nette\Application\UI\Multiplier')->yes()) {
+                // TODO: How to determine component type?
```
lulco: maybe we should collect them as wildcard name subcomponents, I don't use Multiplier so I don't know how it works but from test case I assume we could find all return types of components in Multiplier constructor and use `*` as their name, then we check `*` here in  Visitor.

I'm just not sure why there is not `multiplied` used in latte. How Nette solves the problem which component should be rendered if there are more than one controls in Multiplier?

lulco: ou, it's not array, it's callable factory O:-) Then it is only this one type returned by `createComponentMultiplied`

MartinMystikJonas: Yes Multiplier is just simple object that creates sub-components on demand. But we need to know what type these subcomponents have.

lulco: We know it because factory has some return type, in this case it is Form

MartinMystikJonas:
I implemented it using stub that added generics type to Multiplier. I think we should try to push this stup to phpstan-nette where it is more suitable for wider variaty of use-cases.

src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
```diff
@@ -33,6 +35,16 @@ private function findComponentByName(array $components, string $componentName):
             if (count($componentNameParts) === 0) {
                 return $component;
             }
+            $componentType = $component->getType();
+            if ($componentType instanceof ObjectType && $componentType->isInstanceOf('Nette\Application\UI\Multiplier')->yes()) {
```
lulco: I don’t like that Multiplier is mentioned here in node visitor, it should be generic here. Any other component can beháva like multiplier in a wild. Multiplier components should be just collected in component collector (special one for multiplier).

stubs/Nette/Application/UI/Multiplier.stub
```diff
@@ -0,0 +1,19 @@
+<?php
```
lulco, line 1: Not sure why would we need this if we would have collector

lulco (changes requested):

src/Compiler/NodeVisitor/Behavior/ComponentsNodeVisitorBehavior.php
MartinMystikJonas: Not sure how it would work. How would we pass info that it is multiplier-like component? Another property of Component?

lulco: We will not care about it. As I mention before, we will just create Component with subcomponent with special name. Let’s say `*`. And this name will be checked in NodeVisitor.

stubs/Nette/Application/UI/Multiplier.stub
MartinMystikJonas: Collector see return type of createComponent or expression in return - which is not always expression in which Multiplier is created. I see no other way that having generics for that.

MartinMystikJonas:
I rewritted it to collect subcomponents in collector. I also found out we were unable to evaluate encapsed strings in ValiueResolver so I fixed it.

MartinMystikJonas:
@lulco Can we merge it?

lulco:
> @lulco Can we merge it?

Nope, I have to check it and try my concept I suggested to you.

Also there are several problems solved here, create separate PR for them. Some patril fixed can be merged first.

And first change should be split ComponentCollector to subcollectors.

MartinMystikJonas:
I  am not sure what concept do you mean. I already moved detection of Multiplier to collectors

lulco:
In my head using wildcard for subcomponents ment 2 changes + tests, definitelly not 13 changed files (including behavior)

- add collector for Multiplier
- add checking if component has subcomponent with wildcard name as fallback when component with name is not found, return subcomponent of this type in NodeVisitor. End.

MartinMystikJonas:
Well there were two more changes required to make this work.

1) Small change in Component because now we need a way to addSubpomponents not only to overwrite them. 
2) A way to evaluate component names that are not hardcoded constant strings but contains concatenated or necapsed variables (that is why change to Value resolver and Behavior was needed)

MartinMystikJonas:
I can split resolving of encapsed strings to separate PR. Would it be better?

lulco (approved):
No time to check it now 🤷 tests passed, merge it

MartinMystikJonas:
OK. We can make changes later.

PR #312, opened 2023-02-08
Added support for functions

### LINKED ISSUE #313: Unknown functions
Nette function slice is transformed to php function slice which doesn't exist
Failing test case https://github.com/efabrica-team/phpstan-latte/pull/312

In original compiled file there is:
```
$slicedLocalStrings = ($this->global->fn->slice)($localStrings, 1);
```
This is traformed to:
```
$slicedLocalStrings = \slice($localStrings, 1);
```

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/313


MartinMystikJonas:
No idea. It must nenunintwnded side effect of something. Maybe filter related?

lulco:
Code from Engine::compile (originally used in phpstan-latte):
```php
$code = $this->getCompiler()
    ->setContentType($this->contentType)
    ->setFunctions(array_keys((array) $this->functions))
    ->setFilters($this->filters->_origNames)
    ->setPolicy($this->sandboxed ? $this->policy : null)
    ->compile($tokens, $this->getTemplateClass($name), $comment, $this->strictTypes);
```

Code in phpstan-latte now:
```php
$phpContent = $this->engine->getCompiler()->compile(
    $latteTokens,
    $className,
    $this->generateClassComment($className, $context),
    $this->strictMode
);
```

We are missing content type, functions, filters and also policy

MartinMystikJonas:
Because we use Engine that I thought should handle it internally.

MartinMystikJonas:
Hmm it seems taht for LAtte 2 we would need a quite good amount of magic to make it working

MartinMystikJonas:
I added fix to prevent functions replacing

But we need some way to define types of `$this->global->fn->something` in compiled tempalte otherwise we will have these errors instead:

```
+default.latte:91: Cannot access property $slice on mixed.
+default.latte:91: Trying to invoke mixed but it's not a callable.
```

We probably need to do something like we do for filters also for functions including all places where it is used.

lulco:
Yes we have to do similar functionality like with filters and replace $this->global->fn->something to FilterName::methodName in some NodeVisitor

MartinMystikJonas:
And ideally also collector for manually added functions nad configuration for it same as for filters. BUt I never seen it used in any code so maybe keep it just like low-priority issue for now.

lulco:
self-note: add functions to parameters and parameter schema, add readme etc

lulco:
@MartinMystikJonas basically I copy-pasted filters functionality with some small changes and it works!

MartinMystikJonas (approved):

PR #314, opened 2023-02-09
Fixed filters when 'class::method' syntax is used

MartinMystikJonas:


src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php
```diff
@@ -205,8 +205,12 @@ private function createFilterCallNode(string $filterName, array $args): ?Node
         }
 
         if (is_string($filter)) {
-            $args = $this->updateArgs(ReflectionFunction::createFromName($filter), $args);
-            return new FuncCall(new FullyQualified($filter), $args);
+            if (strpos($filter, '::') === false) {
```
lulco: Use str_contains instead

tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
```diff
@@ -20,4 +25,5 @@ public function objectFilter(string $input): string
 $engine->addFilter('closureWithStringParam', function (string $param) {
 });
 $engine->addFilter('objectFilter', [new Filters(), 'objectFilter']);
+$engine->addFilter('methodStringFilter', Filters::class . '::methodFilter');
```
lulco, line 28: Didn’t even know this is possible :)

lulco (approved):

tests/Rule/LatteTemplatesRule/EngineBootstrap/engine.bootstrap.php
MartinMystikJonas: I found it in one of our projects

PR #317, opened 2023-02-15
Fix for {control $variable}

MartinMystikJonas:


lulco:
What is the code in compiled template? We should transform it if possible. I don’t like ignoring next line - I never used it in php classes (and never will). It would be great remove all its uses

MartinMystikJonas:
This is basically just extension for ignore we already have there but for other branch. Problem is that there is always universal code for component name/object in variable. And because in visitors we have no idea which one it is we have to cover both posibilities. BUt that raises errors on stricter levels.

Generated code now looks like this:

```php
        /** @phpstan-ignore-next-line */ // <-- here we have to ignore Condition is always true/false
        if (\is_object($varControl)) {
            $_tmp = $varControl;
            /** @phpstan-ignore-next-line */
        } else { // <-- here we have to ignore Else branch is not reachable...
            $_tmp = $this->global->uiControl->getComponent($varControl);
        }
        /** @phpstan-ignore-next-line */
        if ($_tmp === null) { // <-- here we have to ignore Condition is always true/false again when $varControl and then $_tmp is never null
            throw new InvalidArgumentException();
        }
```

lulco:
So this is another use case for scope in node visitors. I stop counting them but I think they are at least five or six. We definitelly need it to implement.

MartinMystikJonas:
We basically need a rector-like implementation of node visitors combined with scope resolver. But I am afraid it would not be easy to implement. But we can start by checking how it works in rector.

MartinMystikJonas:
At this point I would merge this as kind of quick-fix and I created issue for implementation of Scope aware node visitors #318

lulco (approved):

PR #319, opened 2023-02-16
Fixed slice stubs

MartinMystikJonas:
Prevents "...no value type specified in iterable type array" on higher levels


lulco:
I've local fix with mixed[] but this is probably better

lulco (approved):

PR #321, opened 2023-02-22
Added missing Multiplier.stub config

MartinMystikJonas:


MartinMystikJonas:
I forgot to add stub to config. It worked on my project because I had it in local config left from initial testing

MartinMystikJonas:
@lulco WTF... somehow I managed to push it both into PR and main. No idea how that happened.

So it is already in main and I am closing this PR. LEt me know if there is anything you would like to change but I think this change is quite straightforward.

lulco (commented):
Approved

PR #323, opened 2023-02-24
Scope aware node visitors

### LINKED ISSUE #318: Scope aware node visitors
In multiple occasions we need Scope with type information to effectively modify compiled template code.

We should check how rector solved this and try to implement similar mechanism.

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/318

Naive implementation of getting types via scope.
Prove of concept in next [PR](https://github.com/efabrica-team/phpstan-latte/pull/324)

### DIFF
--- a/src/Compiler/NodeVisitor/Behavior/ExprTypeNodeVisitorBehavior.php
+++ b/src/Compiler/NodeVisitor/Behavior/ExprTypeNodeVisitorBehavior.php
@@ -11,6 +11,7 @@ trait ExprTypeNodeVisitorBehavior
 {
     public function getType(Node $node): ?Type
     {
-        return $node->getAttribute(ExprTypeNodeVisitorInterface::TYPE_ATTRIBUTE);
+        $type = $node->getAttribute(ExprTypeNodeVisitorInterface::TYPE_ATTRIBUTE);
+        return $type instanceof Type ? $type : null;
     }
 }
--- a/src/Compiler/NodeVisitor/Behavior/ExprTypeNodeVisitorBehavior.php
+++ b/src/Compiler/NodeVisitor/Behavior/ExprTypeNodeVisitorBehavior.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior;
+
+use PhpParser\Node;
+use PHPStan\Type\Type;
+
+trait ExprTypeNodeVisitorBehavior
+{
+    public function getType(Node $node): ?Type
+    {
+        return $node->getAttribute(ExprTypeNodeVisitorInterface::TYPE_ATTRIBUTE);
+    }
+}
--- a/src/Compiler/NodeVisitor/NodeVisitorStorage.php
+++ b/src/Compiler/NodeVisitor/NodeVisitorStorage.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor;
 
+use Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior\ExprTypeNodeVisitorInterface;
 use InvalidArgumentException;
 use PhpParser\NodeVisitor;
 
@@ -19,12 +20,12 @@ final class NodeVisitorStorage
         self::WITH_SCOPE => [],
     ];
 
-    public function addNodeVisitor(int $priority, NodeVisitor $nodeVisitor, bool $withScope = false): void
+    public function addNodeVisitor(int $priority, NodeVisitor $nodeVisitor): void
     {
         if ($priority < 0 || $priority > 10000) {
             throw new InvalidArgumentException('Priority must be set between 0 and 10000');
         }
-        $scope = $withScope ? self::WITH_SCOPE : self::WITHOUT_SCOPE;
+        $scope = $nodeVisitor instanceof ExprTypeNodeVisitorInterface ? self::WITH_SCOPE : self::WITHOUT_SCOPE;
         if (!isset($this->nodeVisitors[$scope][$priority])) {
             $this->nodeVisitors[$scope][$priority] = [];
         }
--- a/src/Compiler/NodeVisitor/NodeVisitorStorage.php
+++ b/src/Compiler/NodeVisitor/NodeVisitorStorage.php
@@ -9,26 +9,34 @@ use PhpParser\NodeVisitor;
 
 final class NodeVisitorStorage
 {
-    /** @var array<int, NodeVisitor[]> */
-    private array $nodeVisitors = [];
+    private const WITHOUT_SCOPE = 'without_scope';
 
-    public function addNodeVisitor(int $priority, NodeVisitor $nodeVisitor): void
+    private const WITH_SCOPE = 'with_scope';
+
+    /** @var array<string, array<int, NodeVisitor[]>> */
+    private array $nodeVisitors = [
+        self::WITHOUT_SCOPE => [],
+        self::WITH_SCOPE => [],
+    ];
+
+    public function addNodeVisitor(int $priority, NodeVisitor $nodeVisitor, bool $withScope = false): void
     {
         if ($priority < 0 || $priority > 10000) {
             throw new InvalidArgumentException('Priority must be set between 0 and 10000');
         }
-        if (!isset($this->nodeVisitors[$priority])) {
-            $this->nodeVisitors[$priority] = [];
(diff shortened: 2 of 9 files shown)


MartinMystikJonas:
I will check later. At first sight it looks good. I like idea of storing only type for relevant expr nodes because it will greatly reduce memory consumption I was afraid skyrocketting by saving scope copies. 

Just one thing if visitor make cahnges to code are types refreshed?

lulco:
> Just one thing if visitor make cahnges to code are types refreshed?

Not yet, it is running just once, plan is to use the types only for some fixing generated code - removing always true / false conditions etc. Not change code in a way where types are changed. If we will need it, we can move creating scope into foreach. But it will slow down the extension again because we also have to store and load content to file and analyse types again and again.

MartinMystikJonas:
What about making scope-type assignment as service. DO it once before visitors are called. And if needed visitor can call this service explicitly to refresh types. But I guess we can do it when we encounter first case where one visitor breaks types for following visitor.

lulco:
Yeah we can change it when it comes useful. For now I would go with this solution

src/Compiler/NodeVisitor/Behavior/ExprTypeNodeVisitorBehavior.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Efabrica\PHPStanLatte\Compiler\NodeVisitor\Behavior;
+
+use PhpParser\Node;
+use PHPStan\Type\Type;
+
+trait ExprTypeNodeVisitorBehavior
+{
+    public function getType(Node $node): ?Type
+    {
```
lulco, line 13: Update as in next PR

src/Compiler/NodeVisitor/NodeVisitorStorage.php
```diff
@@ -9,26 +9,34 @@
...
-    public function addNodeVisitor(int $priority, NodeVisitor $nodeVisitor): void
+    private const WITH_SCOPE = 'with_scope';
+
+    /** @var array<string, array<int, NodeVisitor[]>> */
+    private array $nodeVisitors = [
+        self::WITHOUT_SCOPE => [],
+        self::WITH_SCOPE => [],
+    ];
+
+    public function addNodeVisitor(int $priority, NodeVisitor $nodeVisitor, bool $withScope = false): void
```
lulco: add check if there is interface when it should be

lulco:
Self note: add readme

src/Compiler/NodeVisitor/NodeVisitorStorage.php
MartinMystikJonas: Would not it be better to have two methods or differentiate based on visitor implementing ExprTypeNodeVisitorInterface?

src/Compiler/Postprocessor.php
```diff
@@ -30,26 +37,61 @@ final class Postprocessor
...
+        $phpStmts = $this->findNodes($phpContent);
+
+        $nodeScopeResolver = clone $this->nodeScopeResolver;
+        $nodeScopeResolver->processNodes($phpStmts, $scope, function (Node $node, Scope $scope) {
+            if ($node instanceof Expr && !$node instanceof EncapsedStringPart) {
+                $node->setAttribute(ExprTypeNodeVisitorInterface::TYPE_ATTRIBUTE, $scope->getType($node));
+            }
+        });
+
+        foreach ($this->nodeVisitorStorage->getNodeVisitors(true) as $nodeVisitors) {
```
MartinMystikJonas, line 86: so first all non-scope aware visitors are called and then al scope aware? Is there reason for that?

src/Compiler/Postprocessor.php
```diff
@@ -30,26 +37,61 @@ final class Postprocessor
...
+            return '';
+        }
+        require($compileFilePath); // load type definitions from compiled template
+
+        $scope = $this->scopeFactory->create(ScopeContext::create($realPath));
+
+        $phpStmts = $this->findNodes($phpContent);
+
+        $nodeScopeResolver = clone $this->nodeScopeResolver;
+        $nodeScopeResolver->processNodes($phpStmts, $scope, function (Node $node, Scope $scope) {
```
MartinMystikJonas, line 80: What about moving this to ExprTypeNodeVisitorBehavior and run it on demand - when getType is called for first time - might remove need to do it in many templates at all

src/Compiler/Postprocessor.php
```diff
@@ -30,26 +37,61 @@ final class Postprocessor
...
+    public function postProcess(string $phpContent, Template $template, string $compileFilePath): string
     {
         $phpStmts = $this->findNodes($phpContent);
         foreach ($this->nodeVisitorStorage->getNodeVisitors() as $nodeVisitors) {
             $phpStmts = $this->processNodeVisitors($phpStmts, $nodeVisitors, $template);
         }
 
-        return $this->printerStandard->prettyPrintFile($phpStmts);
+        $phpContent = $this->printerStandard->prettyPrintFile($phpStmts);
+        file_put_contents($compileFilePath, $phpContent);
```
MartinMystikJonas, line 68: Why require is needed here? Would not there be a conflict if version required here and final version is different?

src/Compiler/Postprocessor.php
```diff
@@ -30,26 +37,61 @@ final class Postprocessor
...
+            }
+        });
+
+        foreach ($this->nodeVisitorStorage->getNodeVisitors(true) as $nodeVisitors) {
+            $phpStmts = $this->processNodeVisitors($phpStmts, $nodeVisitors, $template);
+        }
+        $phpContent = $this->printerStandard->prettyPrintFile($phpStmts);
+
+        // update contents
+        file_put_contents($compileFilePath, $phpContent);
```
MartinMystikJonas, line 92: If previous version of class is already required what would happed if it is rewritten here? I would think it should raise error "Class XY already declared"

MartinMystikJonas:
Overall it looks great. I have just some minor questions/proposals (see comments). Good work!

Did you tried what performance/memory impact it would have on some real project?

src/Compiler/Postprocessor.php
lulco: Yes, first we run all visitors which don’t need scope. But main reason is how phpstan scope works. We have to store content to file, load that file and also require it because phpstan want to know all symbols - classes. So we can require this file only once because multiple require on same file will throw error. That's also why we can’t create it as service and run it “on demand”. This is what we have to document - scope aware node visitors can’t add new classes because it will not work then.

src/Compiler/Postprocessor.php
lulco: See comment above

src/Compiler/Postprocessor.php
lulco: Phpstan needs to know all symbols. Final version has to have the same classes as the version at this place

src/Compiler/Postprocessor.php
lulco: We are not requiring the same file twice. Here is just content update

lulco:
> Overall it looks great. I have just some minor questions/proposals (see comments). Good work!

Well I am not completely satisfied. Your proposals are correct, just didn’t work for me as I explained in comments. Maybe there are better solutions I just don’t know how to do it better.
 
> Did you tried what performance/memory impact it would have on some real project?

In few of projects I tried it, it was only few seconds on speed. I didn’t track performance / memory. But it is just something we need so we should accept some little slowing down.

src/Compiler/NodeVisitor/NodeVisitorStorage.php
lulco: Yes I can use the interface as a flag

MartinMystikJonas:
Little slow down is ok. I was afraid that it will take muktiple times more time.

MartinMystikJonas:
Idea: Move additional classes to separate file instead of appending to template file. That way we can only require this additional file once. And then we can analyse different versions of template as many time we need.

lulco:
> Idea: Move additional classes to separate file instead of appending to template file. That way we can only require this additional file once. And then we can analyse different versions of template as many time we need.

For new classes this would work, we can keep it as backup plan. Let's try this for now. We will see if we will need it. For now I'm pretty sure we can create all new classes in non-scope node visitors, and load them only once.

Maybe we will need recreate scope multiple times in the future, we are prepared for that.

lulco:
I'm going to merge this PR now. If we will find some issues with it, we will analyse and rewrite it (or we will go back to previous version without scope)

MartinMystikJonas:
Agreed. We can change it later if needed.

PR #324, opened 2023-02-25
Fixed false positives Method Nette\Forms\Controls\BaseControl::getLabelPart() invoked with 1 parameter, 0 required. using scope

lulco:
POC for using scope types in node visitors

caused some false positive issues in other parts of code, which are now ignored or reported to errors - if this will be acceptable, we can rewrite some parts with using scopes.

Target is to remove all ignore comments which are there basically because we didn't have scope before


src/Compiler/NodeVisitor/Behavior/ExprTypeNodeVisitorBehavior.php
```diff
@@ -11,6 +11,7 @@ trait ExprTypeNodeVisitorBehavior
 {
     public function getType(Node $node): ?Type
     {
-        return $node->getAttribute(ExprTypeNodeVisitorInterface::TYPE_ATTRIBUTE);
+        $type = $node->getAttribute(ExprTypeNodeVisitorInterface::TYPE_ATTRIBUTE);
+        return $type instanceof Type ? $type : null;
```
MartinMystikJonas: Would not MixedType be better here? Of do we need to know when type is for some reason not given?

MartinMystikJonas: As I think about it explicit null is probable better

extension.neon
```diff
@@ -144,6 +145,8 @@ services:
             - addNodeVisitor(200, Efabrica\PHPStanLatte\Compiler\NodeVisitor\ChangeNotNullToEqualsNullNodeVisitor())
             - addNodeVisitor(300, Efabrica\PHPStanLatte\Compiler\NodeVisitor\AddFormClassesNodeVisitor())
             - addNodeVisitor(9900, Efabrica\PHPStanLatte\Compiler\NodeVisitor\CleanupNodeVisitor())
+            # Node visitors with scope
+            - addNodeVisitor(100, Efabrica\PHPStanLatte\Compiler\NodeVisitor\FormDynamicInputsNodeVisitor(), true)
```
lulco: last parameter was removed

tests/config.neon
```diff
@@ -5,19 +5,8 @@ parameters:
     checkExplicitMixed: true
     checkImplicitMixed: true
 
-    # based on phpstan/phpstan-nette
```
lulco: this change has been moved to separate PR

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -230,25 +229,30 @@ public function leaveNode(Node $node)
...
 
-        // dynamic inputs
-        if ($node instanceof Expression && $node->expr instanceof Assign &&
-            ($node->expr->expr instanceof Ternary || ($node->expr->expr instanceof Assign && $node->expr->expr->expr instanceof Ternary))
-        ) {
-            $varName = $this->nameResolver->resolve($node->expr->var);
-            if ($varName === 'ʟ_input' || $varName === '_input') {
-                $node->setDocComment(new Doc('/** @var Nette\Forms\Controls\BaseControl $' . $varName . ' @phpstan-ignore-next-line */'));
-                return $node;
+            // temporary added, should be removed in next PR with whole block
```
lulco, line 247: try to solve this in separate PR and remove from here

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -230,25 +229,30 @@ public function leaveNode(Node $node)
...
-                        if (in_array($this->nameResolver->resolve($node->cond->var), ['ʟ_label', '_label'], true) && $node->cond->expr instanceof MethodCall && $this->nameResolver->resolve($node->cond->expr) === 'getLabel') {
-                            return array_merge([new Expression($node->cond)], $node->stmts);
-                        }
+            if ($node->cond instanceof Assign) {
+                if (in_array($this->nameResolver->resolve($node->cond->var), ['ʟ_label', '_label'], true) && $node->cond->expr instanceof MethodCall && in_array($this->nameResolver->resolve($node->cond->expr), ['getLabel', 'getLabelPart'], true)) {
+                    if (in_array(spl_object_hash($node), $this->possibleAlwaysTrueLabels, true)) {
+                        return array_merge([new Expression($node->cond)], $node->stmts);
                     }
+
+                    // temporary added, should be removed in next PR with whole block
```
lulco, line 238: solve in separate PR

MartinMystikJonas (approved):

lulco:
I have to check this PR, I don't remember why I've added so many new TODOes :)

lulco:
Solved in https://github.com/efabrica-team/phpstan-latte/commit/10394bf6fcadcdfd4d0ea64b55402012e2ade3ae different way

PR #326, opened 2023-02-25
Added phpstan-nette extension to tests

lulco:


lulco:
@MartinMystikJonas could you please check why multiplier with phpstan-nette extensions not working? It works only with php7.4 (or latte 2.x?), otherwise it thinks that type of multiplier component is mixed

lulco:
With newest phpstan-nette, only three tests failed. With older one there were 7 test failed.

code looks the same, I think it is something with stubs

MartinMystikJonas:
I will check later. But my guess is that it can be related to return type extension https://github.com/phpstan/phpstan-nette/blob/1.1.x/src/Type/Nette/ComponentModelDynamicReturnTypeExtension.php

MartinMystikJonas:
My guess was right. It is bug in phpstan-nette. If createComponent* method is missing it forcibly changes return type to mixed instead of using default return type of getComponent. I will make PR with fix to phpstan-nette.

MartinMystikJonas:
Will be fixed by https://github.com/phpstan/phpstan-nette/pull/119

lulco:
Why it works on php 7.4 and all lowest tests than?

MartinMystikJonas:
Not sure maybe this bug was not present in older versions of phpstan-nette.

lulco:
On php 7.4 it is newest version of extension too 🤷

lulco:
Btw lowest and “non-lowest” tests use the same version of extension. Some of them works, some not.

MartinMystikJonas:
Weird

lulco:
@MartinMystikJonas any news here?

MartinMystikJonas:
No response to my PR in phpstan-nette yet. I will try politely remind.

MartinMystikJonas:
Fix was merged into phpstan-nette. I updated required phpstan-nette version to version with fix and rebased onto main. Should be ready to merge

lulco (commented):
Great, we can merge

lulco:
@MartinMystikJonas it looks like it is screwed again in new versions of phpstan-nette extension 😢

MartinMystikJonas:
Ondrej for some reson changed implementation back to forced `mixed`. Not sure why. I asked him in PR comments.

MartinMystikJonas:
It seems it is probably "wont't fix" on Ondřej side. So I think we can either add these arrors to expected errors or ignore them in latte.

lulco:
ok we can ignore them. I was thinking about replacing phpstan-nette's extension in our phpstan-latte but I'm not sure if it would be correct

Issue #266, opened 2023-01-18
Replace BetterReflection with PHPStan ReflectionProvider


MartinMystikJonas:

lulco:
I think we can't do more here. I've replaced all I know

PR #327, opened 2023-03-01
Replace better reflection with PHPStan ReflectionProvider - part 1

lulco:
Ref: https://github.com/efabrica-team/phpstan-latte/issues/266

@MartinMystikJonas 

Replaced BetterReflection with ReflectionProvider in NodeVisitors. I can't find some methods from BetterReflection in ReflectionProvider (e.g. reflectFunction, ReflectionFunction::createFromClosure()),  so they are still used there. Also I'm not sure if this is such valueable, because I had to rewrite few other NodeVisitors, and also ReflectionProvider->getClass()->getMethod() requires Scope as second parameter. I don't see such benefit in this approach. But it is working now so we can merge it if you want.

There are still some classes using BetterReflection, but they were added by you and I am not sure how to replace them.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -129,22 +129,23 @@ services:
     phpstanLatteNodeVisitorStorage:
         factory: Efabrica\PHPStanLatte\Compiler\NodeVisitor\NodeVisitorStorage
         setup:
+            # no scope node visitors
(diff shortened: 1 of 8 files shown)


MartinMystikJonas (approved):

MartinMystikJonas:
We do not need to be strict about removing BetterReflection. Main motivation was 1) performance (reflectionProvider do caching) and 2) PHPStan compatibility - Better reflection does not see information from annotations and PHPStan specific things but only native types.

PR #340, opened 2023-04-09
Changed compiled code for n:tag-if conditions with latte 2 (`$ʟ_if[0]` changed to `$ʟ_if0`)

### LINKED ISSUE #339: href in n:tag-if
Check why code:
```php
if ($ʟ_if[2] = $url) {
    echo '            <a href="';
    echo \Latte\Runtime\Filters::escapeHtmlAttr(\Latte\Runtime\Filters::safeUrl($url));
    echo '">
}
```
causes error
```
Parameter #1 $s of static method Latte\Runtime\Filters::safeUrl() expects bool|float|int|Latte\Runtime\HtmlStringable|Nette\HtmlStringable|string|Stringable, string|null given. 
```

on level 8
In playground it looks ok also on level 9 https://phpstan.org/r/caf16332-e465-448b-9390-ac2c45f2d0db

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/339


MartinMystikJonas:
Maybe change it to `$nullOrUrl && $ʟ_if[0] = $nullOrUrl` to keep original logic? But it probably does matter anyway.

lulco:
condition `$ʟ_if[0]` is used at the end of tag, but $nullOrUrl can be changed in the middle of the block so I shouldn't use it there. Implemented solution works, so I'll keep it as is.

PR #341, opened 2023-04-09
Added support for n:ifcontent

### LINKED ISSUE #338: n:ifcontent
`Parameter #1 $str of function rtrim expects string, string|false given.`

n:ifcontent is compiled as try / catch / finally and in finally there is this code:
```
$ʟ_ifc[1] = \rtrim(\ob_get_flush()) === '';
```
ob_get_flush can return false, so we should cast it to string

level 8

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@
 
 ### Added
 - Support for try / catch in foreach
+- Support for n:ifcontent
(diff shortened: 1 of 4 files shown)


MartinMystikJonas:
LGTM

PR #344, opened 2023-04-11
Improved processing of block's missing parameters

### LINKED ISSUE #343: If no parameters are sent to defined block via include, method call is not transformed
Bug is probably somewhere in RenderBlockNodeVisitor
add test case when no params are sent to block
it causes false negatives when we call methods without parameters

lulco:
- default value is used if it is available
- fixed string default values for blocks
- blocks called with no parameters are also transfered to method call

Resolves https://github.com/efabrica-team/phpstan-latte/issues/343

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased][unreleased]
 
+### Changed
+- Improved processing of block's missing parameters - default value is used if it is available
+
+### Fixed
+- String default values for blocks
+- Blocks called with no parameters are also transfered to method call
+
 ## [0.9.0] - 2023-04-11
 ### Changed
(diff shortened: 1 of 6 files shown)


lulco:
@MartinMystikJonas please, check this on your real world projects. I think I solved all usages except one: if you use parameter without type, missing value is not reported because we are using `new MissingBlockParameter` which is accepted as mixed. I think there is no solution without named parameters from PHP 8 because there is no "none" or "void" type to be used instead, so until we have to support PHP 7.4, we have to live with it :)

MartinMystikJonas:
Tested on our biggest project. It found 20 new bugs and about 10 false positives but all of them were caused by coded added to prevent false positives caused by previous versions with force nullability of parameters. So it looks good to me.

MartinMystikJonas (approved):

PR #347, opened 2023-04-12
Added support for any expression as default value of block parameters

### LINKED ISSUE #345: Check support for constants as default values of block parameters
If it is possible in Nette, add support for it also in this extensiono

### LINKED ISSUE #348: Add support for non-empty-array in {define}

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/345
Resolves https://github.com/efabrica-team/phpstan-latte/issues/348

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "php": ">=7.4 <8.3",
         "ext-json": "*",
         "phpstan/phpstan": "^1.10.6",
-        "phpstan/phpstan-nette": "^1.2.6",
+        "phpstan/phpstan-nette": "1.2.6",
         "latte/latte": "^2.11.6 | ^3.0.4",
         "nette/utils": "^3.2|^4.0",
         "nette/finder": "^2.5|^3.0"
--- a/src/Compiler/NodeVisitor/AddParametersForBlockNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/AddParametersForBlockNodeVisitor.php
@@ -11,7 +11,10 @@ use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\ArrayDimFetch;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\BinaryOp\Coalesce;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Name;
 use PhpParser\Node\Param;
 use PhpParser\Node\Scalar\DNumber;
 use PhpParser\Node\Scalar\LNumber;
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -8,7 +8,7 @@
         "php": ">=7.4 <8.3",
         "ext-json": "*",
         "phpstan/phpstan": "^1.10.6",
-        "phpstan/phpstan-nette": "^1.2.6",
+        "phpstan/phpstan-nette": "1.2.6",
```
lulco, line 11: freezed version of phpstan-nette to working one

src/Compiler/NodeVisitor/AddParametersForBlockNodeVisitor.php
```diff
@@ -83,8 +88,11 @@ public function leaveNode(Node $node): ?array
                     }
                 } elseif ($defaultValue === '[]') {
                     $default = new Array_();
+                } elseif (str_contains($defaultValue, '::')) {
+                    [$className, $constantName] = explode('::', $defaultValue, 2);
+                    $default = new ClassConstFetch(new Name($className), $constantName);
                 } else {
```
lulco: what else it could be?

lulco: what about [1, 2, 3], or ['a', 'b', 'c']

MartinMystikJonas: There could be wide variety of different things and it could be nested. Maybe use php-parser to convert string to nodes?

lulco: great idea i'll try it

lulco: if I analyze only parameter default, I get syntax error when default is for example `[1, 2, 3, 'str']` (it contains comma and because we use comma in pattern so we destroy the array).

So I've tried to analyse whole function (`function myBlock($param1, stdClass[] $param2)`) but then I got error because `stdClass[]` is not supported be PHP so we will lost this information about array types.


Any idea how else we can handle it?

lulco: Ha! I did it O:-) check it please

src/Compiler/NodeVisitor/AddParametersForBlockNodeVisitor.php
```diff
@@ -54,12 +55,42 @@ public function leaveNode(Node $node): ?array
...
+                    // replace something[] to array because it is not supported by php for now
+                    $define = str_replace($type . ' ', 'array ', $define);
+                }
+            }
+
+            // create php code from latte code of define
+            $phpContent = '<?php ' . preg_replace(['/^{define /', '/' . $match['block_name'] . ',? /', '/}$/'], ['function ', $methodName . '(', ') {}'], $define);
+
+            // parse php code
+            $stmts = $this->parser->parseString($phpContent);
```
lulco: Add try catch here. Rather lost parameters of block than whole template

lulco: We can do fallback with types and variables - without defaults / maybe use null expr as default?

composer.json
spaze: What was the reason for the change here? Couldn't find this mentioned in the release notes. Any plan to get it working with the latest phpstan/phpstan-nette again? Thx!

lulco: It is discussed here https://github.com/efabrica-team/phpstan-latte/pull/326#issuecomment-1505761868
phpstan-nette 1.2.6 was last (or only version) where our tests for Multiplier passed. But we will probably ignore those errors. @MartinMystikJonas can you prepare PR and unfreeze phpstan-nette?

lulco: @spaze version 0.11.0 was released with updated phpstan-nette extension

spaze: Thanks!

PR #353, opened 2023-04-18
Fixed dynamic labels in form templates

### LINKED ISSUE #352: Call to an undefined method Nette\Forms\Control::getLabel().
I started getting these errors after composer update on one of our projects where we use form templates.

```
Call to an undefined method Nette\Forms\Control::getLabel().
```

for dynamic labels:

```latte
{label $name /}
```

Compiled template looks like this:

```php
echo $ʟ_label = \Nette\Bridges\FormsLatte\Runtime::item($name, $this->global)->getLabel();
```

Any idea why this did not showed earlier? Is it something we changed our side or some change in latte?

MartinMystikJonas:
Resolves #352

### DIFF
--- a/src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
@@ -266,6 +266,7 @@ final class AddFormClassesNodeVisitor extends NodeVisitorAbstract implements For
          * </code>
          */
         if ($node instanceof Echo_ && ($node->exprs[0] ?? null) instanceof MethodCall) {
+            $varName = 'ʟ_input';
             /** @var MethodCall $methodCallExpr */
             $methodCallExpr = $node->exprs[0];
             $methodCallVar = $methodCallExpr->var;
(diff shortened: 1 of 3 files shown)


src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -266,6 +266,7 @@ public function leaveNode(Node $node)
          * </code>
          */
         if ($node instanceof Echo_ && ($node->exprs[0] ?? null) instanceof MethodCall) {
+            $varName = 'ʟ_input';
```
MartinMystikJonas, line 269: Moved to same place to keep it similar to following code

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/Fixtures/templates/Forms/default.latte
```diff
@@ -1,7 +1,7 @@
 {form firstForm}
     {input text} <label n:name="text">Text</label><input n:name="text">
     {input textarea id => "myId"}
-    {input password}
+    {input password}{label password /}
```
lulco: I think we have covered this, see line 12, 15 etc

lulco (approved):

PR #357, opened 2023-04-24
Line of compilation error (if available) instead of -1

### LINKED ISSUE #356: Try to handle error line for compilation errors
Now, if there is some exception, we use line -1. Check if there is any chance to identify real latte line

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/356

### DIFF
--- a/src/Compiler/Compiler/Latte2Compiler.php
+++ b/src/Compiler/Compiler/Latte2Compiler.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Efabrica\PHPStanLatte\Compiler\Compiler;
 
+use Latte\CompileException;
 use Latte\Engine;
 use Latte\Runtime\Defaults;
 use Latte\Runtime\FilterExecutor;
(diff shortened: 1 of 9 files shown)


src/Compiler/Compiler/Latte2Compiler.php
```diff
@@ -43,16 +44,27 @@ protected function createDefaultEngine(): Engine
...
+            $phpContent = $compiler
+                ->setFunctions(array_keys($this->getDefaultFunctions()))
+                ->compile(
+                    $latteTokens,
+                    $className,
+                    $this->generateClassComment($className, $context),
+                    $this->strictMode
+                );
+        } catch (CompileException $e) {
+            $line = isset($latteTokens)
```
lulco, line 61: inspired by latte 2 Engine

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -825,6 +826,11 @@ public function testStartupView(): void
     public function testSnippets(): void
     {
         $this->analyse([__DIR__ . '/Fixtures/SnippetsPresenter.php'], [
+            [
+                'Combination of n:snippet with n:foreach is invalid, use n:inner-foreach ' . (LatteVersion::isLatte3() ? '(on line 3 at column 6)' : '(on line 3)'),
```
lulco, line 830: messages are different in latte 2 and latte 3, but I think it is good to know here the error raised so I keep it there.

MartinMystikJonas (approved):

PR #358, opened 2023-04-24
Unfreezed phpstan-nette extension

lulco:
@MartinMystikJonas can you please handle these errors? I don't know if we should just ignore them, or we can create ReturnTypehint extension here in our extension and not use phpstan-nette's one.


MartinMystikJonas:
I think we can ignore this exact error in all latte templates

lulco:
OK I'm just afraid we will ignore errors not associated to Controls. But we can try

PR #365, opened 2023-05-05
Collect renders also from calls to Latte\Engine

MartinMystikJonas:
Related to Slim framework use-case where renders are triggered by call to `$engine->renderToString()`

### DIFF
--- a/src/LatteContext/Collector/TemplateRenderCollector.php
+++ b/src/LatteContext/Collector/TemplateRenderCollector.php
@@ -23,6 +23,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\ThisType;
+use PHPStan\Type\TypeCombinator;
 
 /**
  * @extends AbstractLatteContextCollector<CollectedTemplateRender|CollectedError>
@@ -137,7 +138,12 @@ final class TemplateRenderCollector extends AbstractLatteContextCollector
         }
 
         $calledType = $scope->getType($node->var);
-        if (!$this->templateTypeResolver->resolve($calledType)) {
+        if ($calledType instanceof ThisType) {
+            $calledType = $calledType->getStaticObjectType();
+        }
+        $calledType = TypeCombinator::removeNull($calledType);
+        $engineType = new ObjectType('Latte\Engine');
+        if (!$this->templateTypeResolver->resolve($calledType) && !$engineType->isSuperTypeOf($calledType)->yes()) {
             return null;
         }
 
--- a/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Resolve/SomeControl.php
+++ b/tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Resolve/SomeControl.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\SimpleControl\Fixtures\Resolve;
 
 use Exception;
+use Latte\Engine;
 use Nette\Application\UI\Control;
 
 final class SomeControl extends Control
@@ -204,5 +205,11 @@ final class SomeControl extends Control
     private function neverReturn(): void
     {
         die();
+    }
+
+    public function renderRenderEngine(): void
+    {
+        $engine = new Engine();
+        $engine->render(__DIR__ . '/engine.latte', ['a' => 'a', 'b' => 'b']);
     }
 }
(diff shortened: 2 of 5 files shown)


lulco:
can you provide some test case?

src/LatteContext/Collector/TemplateRenderCollector.php
```diff
@@ -137,7 +137,11 @@ public function processNodeInPhp(Node $node, Scope $scope): ?array
         }
 
         $calledType = $scope->getType($node->var);
-        if (!$this->templateTypeResolver->resolve($calledType)) {
+        if ($calledType instanceof ThisType) {
+            $calledType = $calledType->getStaticObjectType();
+        }
+        $engineType = new ObjectType('Latte\Engine');
+        if (!$this->templateTypeResolver->resolve($calledType) && $engineType->isSuperTypeOf($calledType)->no()) {
```
lulco: ->no() means it is maybe, I prefer to use `! ->yes()`

src/Resolver/TypeResolver/TemplateTypeResolver.php
```diff
@@ -18,7 +18,9 @@ final class TemplateTypeResolver
     public function resolve(Type $type): bool
     {
         if ($type instanceof ObjectType) {
-            return $type->isInstanceOf('Nette\Application\UI\Template')->yes() || $type->isInstanceOf('Nette\Application\UI\ITemplate')->yes();
+            return $type->isInstanceOf('Nette\Application\UI\Template')->yes() ||
+                $type->isInstanceOf('Nette\Application\UI\ITemplate')->yes() ||
+                $type->isInstanceOf('Latte\Runtime\Template')->yes();
```
lulco: what will change this? We already test rendered templates

MartinMystikJonas: Use referenced `Nette\Application` tempaltes not `Latte\Runtime` created by `$engine->createTempalte()`

src/LatteContext/Collector/TemplateRenderCollector.php
MartinMystikJonas: I think we want to cover maybes too. If render is called on union type that includes `Latte\Engine`

MartinMystikJonas:
I will write test

src/LatteContext/Collector/TemplateRenderCollector.php
lulco: I think we should be consistent, so for now use ! yes as it is everywhere

src/Resolver/TypeResolver/TemplateTypeResolver.php
lulco: I see, I never use it this way

MartinMystikJonas: Me neither but it can be used this way

src/LatteContext/Collector/TemplateRenderCollector.php
MartinMystikJonas: ok

MartinMystikJonas:
Updated and test added

src/LatteContext/Collector/TemplateRenderCollector.php
```diff
@@ -137,7 +138,12 @@ public function processNodeInPhp(Node $node, Scope $scope): ?array
         }
 
         $calledType = $scope->getType($node->var);
-        if (!$this->templateTypeResolver->resolve($calledType)) {
+        if ($calledType instanceof ThisType) {
+            $calledType = $calledType->getStaticObjectType();
+        }
+        $calledType = TypeCombinator::removeNull($calledType);
+        $engineType = new ObjectType('Latte\Engine');
```
lulco, line 145: just question, why we use Latte only here? It could be in TemplateTypeResolver.php with other classes, or not?

tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Resolve/SomeControl.php
```diff
@@ -205,4 +206,10 @@ private function neverReturn(): void
     {
         die();
     }
+
+    public function renderRenderEngine(): void
```
lulco, line 210: Will this work when it will be in some custom Class and not in Control / Presenter?

src/LatteContext/Collector/TemplateRenderCollector.php
MartinMystikJonas: We use templateTypeResolver in different context that are not valid for Engine

tests/Rule/LatteTemplatesRule/SimpleControl/Fixtures/Resolve/SomeControl.php
MartinMystikJonas: It would as long as that custom class/method is matched by some TemplateResolver

src/LatteContext/Collector/TemplateRenderCollector.php
lulco: I am pretty sure it is not valid for Latte\Runtime\Template also

MartinMystikJonas: Did not thing of that. I will check.

lulco (approved):
Check it and we can try 🤷‍♂️

src/LatteContext/Collector/TemplateRenderCollector.php
MartinMystikJonas: You are right LAtte\Runtime\Template acnnot be used this was and calling render on it is marked as internal so I think we can ignore this use-case a cover only Engine.

PR #367, opened 2023-05-11
TemplateRender subcollectors

MartinMystikJonas:
Support for subcollector for TemplateRender. Sub collectors can return CollectError now (needed by TempalteRender subcollector but I make it supported by all subcollectors). Also added few helper methods for repeated code.

### DIFF
--- a/src/LatteContext/LatteContextHelper.php
+++ b/src/LatteContext/LatteContextHelper.php
@@ -5,17 +5,25 @@ declare(strict_types=1);
 namespace Efabrica\PHPStanLatte\LatteContext;
 
 use Efabrica\PHPStanLatte\Template\Variable;
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\Variable as VariableExpr;
+use PHPStan\Analyser\Scope;
 use PHPStan\Type\Constant\ConstantArrayType;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ObjectType;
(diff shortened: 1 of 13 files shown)


src/LatteContext/LatteContextHelper.php
```diff
@@ -44,4 +52,61 @@ public static function variablesFromTemplateType(string $class): array
...
+            return [];
+        }
+
+        return LatteContextHelper::variablesFromType($scope->getType($expr));
+    }
+
+    /**
+     * @param class-string|class-string[] $classes
+     */
+    public static function isClass(Node $node, Scope $scope, $classes): bool
```
lulco, line 71: I'm not sure if this is only Latte context thing. Maybe it should be part of resolvers (like we have TemplateTypeResolver), or even better use some built-in function from phpstan?

lulco:
It's kinda weird when we return Error in response, I really don't like this approach. Isn't there any better approach?

MartinMystikJonas:
We already use same approach in normal Latte cantext collectors. Only other way would be to create result classes for each collector/subcollector that holds both results and errors like we did in template resolvers. But it would overcomplicate things IMHO. Or limit errors to only one per collector/subcollector and remove posibility to return some results alongside errors and use exception. But that would greatly limit our options.

src/LatteContext/LatteContextHelper.php
MartinMystikJonas: Cirrently it is only latte context thing but conceptually it is not. We could either create some general helper for types in root namespace or create NodeClassResolver.

MartinMystikJonas: I think there is no such finction in phostan because this helper encapsulates some repated but specific logic.

lulco (approved):

PR #372, opened 2023-05-18
Changed types handling - used smart extract feature from PHPStan

### LINKED ISSUE #370: Array shape with optional parameters thinks that they are always set
```php
/** @var $params array{command?: string} */
$this->template->setParameters($params);
```

```latte
Command {$command ?? ''}
```

Error: `Variable $command on left side of ?? always exists and is not nullable.`

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/370


src/LatteContext/CollectedData/CollectedVariable.php
```diff
@@ -63,12 +63,12 @@ public function isDeclared(): bool
         return $this->declared;
     }
 
-    public static function build(Node $node, Scope $scope, string $name, Type $type, bool $declared = false): self
+    public static function build(Node $node, Scope $scope, string $name, Type $type, bool $declared = false, bool $mightBeUndefined = false): self
```
lulco, line 66: @MartinMystikJonas do you remember what is $declared for? There is no doc for that

src/Template/Variable.php
```diff
@@ -15,10 +15,13 @@ final class Variable implements NameTypeItem, JsonSerializable
...
     private Type $type;
 
-    public function __construct(string $name, Type $type)
+    private bool $mightBeUndefined;
+
+    public function __construct(string $name, Type $type, bool $mightBeUndefined = false)
     {
         $this->name = $name;
         $this->type = $type;
+        $this->mightBeUndefined = $mightBeUndefined;
```
lulco, line 24: Once it might be use when variable is assigned in condition in presenter like:
```php
if ($someCondition) {
    $this->template->a = 'abcd';
}
```
We just have to resolve if variable is assigned in all branches of the method or not, but may be there is some logic in phpstan itself, we shouldn't reinvent the wheel.

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
```diff
@@ -316,6 +316,26 @@ public function testVariables(): void
                 3,
                 'differentRenderConditional.latte',
             ],
+            [
+                'Dumped type: mixed', // TODO should be string, but PHPStan resolves this as mixed for now
```
lulco: https://phpstan.org/r/c5055b61-4862-4f5b-bfb8-7d570e68fb73

src/LatteContext/CollectedData/CollectedVariable.php
MartinMystikJonas: It is used to differentiate between discovered types and explicitly declared types (using @phpstan-latte-* annotations) whoch have higher priority when types are conflicting

lulco: Oh, I see. OK, so I'm not duplicating functionality here

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
lulco: After discussion with @ondrejmirtes, issue has been created https://github.com/phpstan/phpstan/issues/9331, maybe there is a light at the end of the tunnel

MartinMystikJonas: This could be source of unexpected sideeffects. I would wait for solving it in PHPStan before merging this.

src/Template/Variable.php
MartinMystikJonas: Not sure if it is possible to solve. We encountered similar problem before with terminating methods and did not find a solution. In my projects I solved it by using temp variable and assigning to template outside condition.

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
lulco: Well yes, with mixed type it could be. On the other hand, the most questions I get are about “variable is always defined” error 😂

MartinMystikJonas: Well at this point it is choosing between two non-ideal options. But loosing type of all possibly undefined variables is IMHO worse. Maybe ignore "always defined" errors as temporary solution?

lulco: I would like to not ignore errors at all. This error is fixable with adding variable to each branch of program. I'll mark this as draft while there will be some support from phpstan itself

src/Template/Variable.php
MartinMystikJonas: Note: It would also need to cover cases when variable is assigned in conditionaly called method/function.

lulco:
@MartinMystikJonas any idea why
we have still $a as mixed?

```
-arrayShapeParams.latte:03: Dumped type: string
+arrayShapeParams.latte:03: Dumped type: mixed
```

in compiled template there is this:
```
/** @var array{baseUrl: string, basePath: string, ʟ_fi: Latte\Runtime\FilterInfo, ʟ_tag: array<string>, ʟ_if: array, ʟ_ifc: array, ʟ_try: array, ʟ_loc: array, ʟ_tmp: mixed, ʟ_input: Nette\Forms\Controls\BaseControl, ʟ_label: (Nette\Utils\Html | string | null), user: Nette\Security\User, startup: 'startup', startupParent: 'startupParent', presenter: Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\PresenterWithoutModule\Fixtures\VariablesPresenter, control: Efabrica\PHPStanLatte\Tests\Rule\LatteTemplatesRule\PresenterWithoutModule\Fixtures\VariablesPresenter, flashes: array<stdClass>, a?: string, b: int, c: (string | null)} $__variables__ */
        \extract($__variables__);
```

If I use the same array shape in phpstan.org/try I get correct type:
https://phpstan.org/r/cc705c48-09f5-4eff-902f-4ce84eafbaf8

lulco:
hmm but if I use array shape as @var it is same as in app, so I think there is maybe som bug in phpstan itself?

https://phpstan.org/r/8562ce02-9e8b-473f-8904-cd6f00a9dffa

lulco:
What about to do this:

```php
/** @var arrayShape $__variables */
$__variables = $this->params;
extract($__variables);
``` 

it should do only one extract and should work as expected

lulco:
Confirmed: https://phpstan.org/r/bc248c3f-893d-43a0-a926-cd3ab2237a93

src/Error/Transformer/UndefinedVariableTransformer.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Efabrica\PHPStanLatte\Error\Transformer;
+
+use Efabrica\PHPStanLatte\Error\Error;
+
+final class UndefinedVariableTransformer implements ErrorTransformerInterface
```
lulco: After using extract of array shape, not defined variables are marked with error `Undefined variable ...`. This transformer serves for back compatibility, but I would like to make difference between `undefined` and `might not be defined` variables - what do you think? Should we keep this transformer or make BC break?

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
lulco: I think I solved this with few extract calls for each type of variables (variables, default variables, filters, functions) instead of `@var` annotations. Everything works as expected in tests, need some testing in real world apps. Maybe you could check this branch too

src/Template/Variable.php
lulco: I said it will be possible mar variable as optionally defined :) I just don't know how to achieve it :))

src/Error/Transformer/UndefinedVariableTransformer.php
MartinMystikJonas: I would go for BC break and more specific message

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
MartinMystikJonas: I will check next week. I will be without computer for rest of the week.

src/Error/Transformer/UndefinedVariableTransformer.php
lulco: OK transformer removed and changed errors to undefined variable

tests/Rule/LatteTemplatesRule/PresenterWithoutModule/LatteTemplatesRuleForPresenterTest.php
lulco: I've removed transformer and changed error messages. If you don't have time, I'd merge it

MartinMystikJonas: I am back home but with bad cough and on top of that eye infection :-( So I do not feel for doing reviews. if tests pass then you can merge it.

PR #376, opened 2023-05-25
Multi registered components

### LINKED ISSUE #377: No subcomponents found if there is component registered multiple times
If component of some type and same name is registered more than once, we somehow lost its subcomponents
```
if ($someCondition) {
    $presenter->addComponent(new SomeControl(), 'someControl');
} else {
    $presenter->addComponent(new SomeControl(), 'someControl');
}
```

Failing test case https://github.com/efabrica-team/phpstan-latte/pull/376

I think there is some problem in ItemCombinator or ComponentFinder (more probably ItemCombinator).

@MartinMystikJonas any idea what it can be?

maybe overwriting component when the name is the same:
```
foreach ($itemArray as $item) {
    $itemName = $item->getName();
    $merge[$itemName] = $item;
}
```
?

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/377


lulco (commented):
It works great! thanks

PR #378, opened 2023-05-26
Separated collection of Form Containers

lulco:
We separate collecting Form Fields and Form Containers so we can create fake classes for each Container (like we do for Forms), and these classes can be used in Form::offsetGet() conditional return type.

It means:
```latte
{$form[container][field]}
```

will be BaseControl instead of IComponent

There will be some other changes because of renaming *Field classes to *Control, but I would like you to check it now when there are only 6 changed files without renaming :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,13 @@
 
 ### Changed
 - Separated collection of Form Containers
+- Renamed CollectedFormField to CollectedFormControl (**BC break**) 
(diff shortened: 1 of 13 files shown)


MartinMystikJonas (approved):

MartinMystikJonas:
Real cahllenge would be to remember/find out name of container to which control are added (especially when it would ba added in some deeper call cahin). It could by done using custom types but it is a bit complicated. But this change is good first step in that direction.

PR #379, opened 2023-06-02
Added support for numeric form container names

lulco:


src/Compiler/NodeVisitor/ContainerArrayFetchChangeIntegerToStringNodeVisitor.php
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        $varType = $this->getType($node->var);
+        if ($varType === null) {
+            return null;
+        }
+
+        $dimType = $node->dim ? $this->getType($node->dim) : null;
+        if ((new ObjectType('Nette\Forms\Container'))->isSuperTypeOf($varType)->yes() && $dimType instanceof ConstantIntegerType) {
+            $node->dim = new String_((string)$dimType->getValue());
```
MartinMystikJonas, line 33: Why is this type change needed?

src/Resolver/ValueResolver/ValueResolver.php
```diff
@@ -134,6 +150,22 @@ public function resolveStrings(Expr $expr, Scope $scope): ?array
         if ($values === null) {
             return null;
         }
+
         return array_filter($values, 'is_string');
     }
+
+    /**
+     * @return (string|int)[]|null
```
MartinMystikJonas: `array<int|string>` is more readable than `(int|string)[]` I think

src/Compiler/NodeVisitor/ContainerArrayFetchChangeIntegerToStringNodeVisitor.php
lulco: Because getValue() returns integer here and String_ requires string

src/Resolver/ValueResolver/ValueResolver.php
lulco: I can change it

src/Compiler/NodeVisitor/ContainerArrayFetchChangeIntegerToStringNodeVisitor.php
MartinMystikJonas: I mean why is it even needed to convert it to _String

lulco: I see :) sorry. Here is the thing:
Containers can be called with integers. We have out containers with string names. In nette they are converted to string in anyway when getting component with name. But in compiled php there is
```php
$something = $form[0];
```

which causes phpstan error “Offset 0 doesn't exist” so I just replaced it with 

```php
$something = $form['0'];
```

and now it works without any error

MartinMystikJonas: I see.

MartinMystikJonas (approved):

PR #397, opened 2023-07-14
Added int to Runtime.stub

### LINKED ISSUE #396: Form containers with numeric name
First, I know that form containers may not be fully supported yet, see #380, but this may or may not be directly related so I'm still filing this.

I have a form with multiple containers (they're nested but that's unrelated I think)

This is how I build the containers, notice the `for` loop and `addContainer($i)`:
```php
		$disclosureNewContainer = $disclosureContainer->addContainer('new');
		for ($i = 0; $i < $newDisclosures; $i++) {
			$disclosureNewCountContainer = $disclosureNewContainer->addContainer($i);
```

In a template:
```latte
	{formContainer new}
		{formContainer $i}  <-- line 85
```

This is how it's compiled by the extension:
```
             /* line 84 */
            $this->global->formsStack[] = $formContainer = $form['new'];
            /* line 85 */
            $this->global->formsStack[] = $formContainer = \Nette\Bridges\FormsLatte\Runtime::item($i, $this->global);
```

I get this error on level 5:
```
 ------ ---------- ---------------------------------------------------------------------------------------------------------------------------------------------
  Line   Compiled   Admin/Presenters/templates/Pulse/passwordsStorages.latte rendered from MichalSpacekCz\Admin\Presenters\PulsePresenter::passwordsStorages
         line       See compiled template: /tmp/phpstan-latte/app/Admin/Presenters/templates/Pulse/passwordsStorages.latte.8f4b78add10adffe699f7ad521d7e14e.php
 ------ ---------- ---------------------------------------------------------------------------------------------------------------------------------------------
  85     437        Parameter #1 $item of static method Nette\Bridges\FormsLatte\Runtime::item() expects object|string, int<0, 2> given.
 ------ ---------- ---------------------------------------------------------------------------------------------------------------------------------------------
```

I'm not sure where `object|string` is coming from, because `\Nette\Bridges\FormsLatte\Runtime::item($item, $global): object` doesn't use any type declarations at all: https://github.com/nette/forms/blob/8a23250f05443cae8b7c8294003c3f2a9d38ffc5/src/Bridges/FormsLatte/Runtime.php#L115

And eventually does this
```php
return $form[$item];
```

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 - Changed types handling - used smart extract feature from PHPStan
 - Not defined variables are marked with error `Undefined variable ...` (**BC break** - if error was ignore, you need to change ignored error pattern)
+- Added type int as param for Runtime::item() method
 
 ### Fixed
 - Collecting of conditionally defined variables from array
--- a/src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
+++ b/src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
@@ -134,8 +134,8 @@ final class AddFormClassesNodeVisitor extends NodeVisitorAbstract implements For
                     $itemArgument = $node->getArgs()[0] ?? null;
                     $itemArgumentValue = $itemArgument ? $itemArgument->value : null;
 
-                    if ($itemArgumentValue instanceof String_) {
-                        $controlName = $itemArgumentValue->value;
+                    if ($itemArgumentValue instanceof String_ || $itemArgumentValue instanceof LNumber) {
+                        $controlName = (string)$itemArgumentValue->value;
                         // TODO remove when container are supported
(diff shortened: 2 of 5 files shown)


lulco:
@spaze check this… but maybe there will be some other issue later, because we are returning BaseControl here and in your case it is Container.

spaze:
Can confirm it fixes the issue, and can confirm that on a higher level the BaseControl is causing some headaches :-)

spaze:
Thanks!

lulco:
Yeah we had headaches with original type IComponent and we decided to use BaseControl instead because it was BaseControl in 90+% of usage… but forms are really complex, we need to do more work with them.

lulco:
TODO add test case and update changelog

PR #429, opened 2023-08-08
Added feature phpstanCommand to analyse compiled templates via separated command

lulco:


MartinMystikJonas:
I will try it on our projects.

Just few questions: 

1) Why is it required to specify phpstan commamd manually? I think it should be possible to get original command used to run phpstan automatically and reuse it. But it might be good idea to allow running it with different config (to allow customization of rules just for latte) 
2) Would not this run latte context collecting rules again on collectedPaths? I think we should skip our rule when analysisng compiled templates
3) It would be good idea to change tmp dir to avoid invalidating of all PHPStan caches (it is invalidated if phpstan is run with different params/target)

lulco:
> I will try it on our projects.
> 
> Just few questions:
> 
> 1. Why is it required to specify phpstan commamd manually? I think it should be possible to get original command used to run phpstan automatically and reuse it. But it might be good idea to allow running it with different config (to allow customization of rules just for latte)

It was my first thought, but I wasn't able to get actual running command in our phpstan latte rule. Especially in tests where no vendor/bin/phpstan is running (it is probably vendor/bin/phpunit there).
But as you said, now you can specify some other phpstan command with different config, so it is feature :D

> 2. Would not this run latte context collecting rules again on collectedPaths? I think we should skip our rule when analysisng compiled templates.

In my first tests it works ok, but today I tried it on bigger project and I got infinite loop. But I really don't understand why it happens. I have to check. Maybe it is collectedPaths as you said.

> 3. It would be good idea to change tmp dir to avoid invalidating of all PHPStan caches (it is invalidated if phpstan is run with different params/target)

I think path is part of phpstan cache key so it will change becasuse {dir} is changed on every run. But we can check this.

lulco:
ad 2. I found bug in my custom collector where I always add some classes to analyse, I fixed it so my infinite loop disappeared

lulco:
I did some measurement on bigger projects and I again achieved 12% speed up. But it was change from 5:37 to 4:55 so it is still little bit slow :(

MartinMystikJonas:
"changing the list of analysed paths between runs actually invalidates the whole result cache" https://phpstan.org/blog/why-you-should-always-analyse-whole-project

MartinMystikJonas:
I think we should made use of results cache for analysising compiled templates. It can substatntially increase speed of repeated runs. 

We should have
- separate tmp dir for template run and from project run
- we should avoid changing path/mtime/content of compiled templates - if it does not change then it would not be analysed agin every time but cached result would be used

lulco:
I usually delete whole cache before phpstan run because when I change some vendor files, phpstan not reflect it (because of cache). I would like to speed up first run. Repeated runs are low priority for me here. I have projects with run more then 10 minutes.

jakubvojacek:
>  It can substatntially increase speed of repeated runs.

repeated runs is mostly what we use, we even save cache in CI runners to speed up subsequential runs. The slowness is the main reason why we didnt start using phpstan-latte yet. Right now, phpstan takes around 3-4 seconds to complete on my work machine which is a great time as I like to run it very often

MartinMystikJonas:
I also use mostly repeated runs. I think @lulco can focus on speed up of first run and I will look at speed up of repeated runs.

lulco:
What about to create “shared” dir for compiled templates? You can use it in your CI. We will check template in this dir and if it exists we just copy it to target dir. So in repeated runs we can benefit from it - no postprocessing will be executed because compiled template will be ready to analyse, right?

lulco:
@jakubvojacek welcome back :)

> The slowness is the main reason why we didnt start using phpstan-latte yet. Right now, phpstan takes around 3-4 seconds to complete on my work machine which is a great time as I like to run it very often

How long phpstan latte takes? How many templates do you have? Can you share some of your projects?

My first target was to make it work, most of my projects takes max 1 minute, but I also have some which takes 10 minutes, so yes, now I want to speed it up :)

jakubvojacek:
> How long phpstan latte takes? How many templates do you have? Can you share some of your projects?

about 1-2 minutes if I recall but its been a few months since I tried last time. We don't really have many templates and they're rather non-complicated, I believe. I only work on one project (one project, 13 years, a long time :)) but I cannot share it. 

I just did some basic math via `ls`, `grep` and `wc -l`, looks like 410 latte files in total

MartinMystikJonas:
I finally tested it and I got quite good speed up on our biggest project. We fot from 4:25 to 3:20 (-25%). Good work!

docs/features.md
```diff
@@ -0,0 +1,42 @@
...
+            transformDynamicFormControlNamesToString: true    
+```
+
+## phpstanCommand
+
+- type: nullable string
+- default: null
+
+Experimental feature which should speed up latte templates analysis. It executes defined command via shell exec and collect errors from this separated run.
+As there is not possible to know the name of the directory where compiled templates are stored, you can use `{dir}` placeholder. We also add `--error-format json` to the end of command so you can't use your error formatter. But you can use any other phpstan command options like `configuration`, `level` etc.
```
lulco, line 27: It seems you can't use debug either. It's probably part of output which is json decoded

lulco: Also mention we can use it without —no-progress so we will see progress of analysing templates

MartinMystikJonas: Maybe it would be better to configure only customisable parts like binPath, configPath, level,... And create command ourrselves. WDYT?

lulco: Not sure, its too many configs :) I like the string. It  just has some limitations which should be documented

MartinMystikJonas: Well my idea was to keep it on bare minimum. Almost everything relevant could be set in custom config file

MartinMystikJonas:
PHPStan itslef uses autodetection of executable path for spawning paralel workers. See https://github.com/phpstan/phpstan-src/blob/8b0bfd44f60a092c836570081228abc28566e1b1/src/Command/AnalyserRunner.php#L53

I does not work in tests but should work in real app. So maybe we should autodetect this was if command is not exolicitly provided and in tests set it manually.

lulco:
> PHPStan itslef uses autodetection of executable path for spawning paralel workers. See https://github.com/phpstan/phpstan-src/blob/8b0bfd44f60a092c836570081228abc28566e1b1/src/Command/AnalyserRunner.php#L53
> 
> I does not work in tests but should work in real app. So maybe we should autodetect this was if command is not exolicitly provided and in tests set it manually.

Yes, but it executes the same command in parallel workers, we want to run some other command (at least target path is different).

Let's focus on collecting context (could it be collected in parallel?) and maybe fast check in node visitors (could be latte files compiled and modified in parallel?)

MartinMystikJonas:
I think we can construct some default command based on argv and allow user to supply custom command.

Context collection and visitor would be problem to run in parallel because of problems with type serialization we encountered before.

Compilation itself (without visitors) can be run in parallel without problems I think.

And for visitors we convert types to phpdoc later anyway so maybe do it before passing it to parallel workers.

PR #433, opened 2023-08-22
Added testing layout files for each presenter's action

### LINKED ISSUE #424: Test @layout.latte
Add tests for @layout.latte files - we need to test them with all presenter / action contexts
Do not forget that template can change layout

lulco:
Resolves https://github.com/efabrica-team/phpstan-latte/issues/424


lulco:
@MartinMystikJonas Hi, I would need some help here from you. I don’t understand why the layout isn't tested with all presenter actions. Where should I add this new functionality? I add it to Nette UI presenter and also to presenter standalone templates but it seems it is not everything :) I get lost so I just pushed what I have for now - I didn’t finished tests (expected errors) for now because I'm not sure what is correct :)

MartinMystikJonas:
I will check later today or tommorow. Which actions are skipped?

lulco:
For example Variables:arrayShape, objectShape etc. It covers only four or five actions from Variables presenter

I run just that one test (LatteTemplatesRuleForPresenterTest) with `- -filter Variables`

lulco:
Ha! I finally found it :)

```
if (!array_key_exists($templatePath, $alreadyAnalysed)) {
    $alreadyAnalysed[$templatePath] = 1;
} elseif ($alreadyAnalysed[$templatePath] <= 3) {
    $alreadyAnalysed[$templatePath]++;
} else {
    continue; // stop recursion when template is analysed more than 3 times in include chain
}
```

This caused layout.latte is analysed only 4 times :) then it is stopped

MartinMystikJonas:
Hmm this is weird. That recursion failsafe should trigger only in nested include calls. Each root template (action) should be serparate here so count layout only once.

MartinMystikJonas:
I will try to find some time to check it later today. It seem like we might have bug in that recursion failsafe. And it could cause skipping analysis in other cases as well.

lulco:
Yes it can, for example if we have some template used in different components it is tested only for four of them

MartinMystikJonas:
Then it does not work as originally intended. Original intention was to prevent infinite recursion with includes A  > B > A > B > A > B > Now A would be fourth time in stack and it should be stopped.

So I think we should revert your second commit and fix it in different means to work as originally intended. I can look at it.

lulco:
> So I think we should revert your second commit and fix it in different means to work as originally intended. I can look at it.

Well, do not revert whole second  commit only that isLayout flag :)

MartinMystikJonas:
Could you revert all related code to isLayout flag? I will prepare separate PR with fix for already Analaysed

lulco:
> Could you revert all related code to isLayout flag? I will prepare separate PR with fix for already Analaysed

Sure, I’ll prepare it later today or tomorrow.

lulco:
> Could you revert all related code to isLayout flag? I will prepare separate PR with fix for already Analaysed

Great, please add test case I mentioned before - shared template for multiple (more than 4) Controls / actions

MartinMystikJonas:
I already have workign fix and we already have such test case there (in VariablesPresenter) - we just had wrong expectations in tests. I am fixing tests now

src/Resolver/LayoutResolver/LayoutPathResolver.php
```diff
@@ -0,0 +1,31 @@
...
+final class LayoutPathResolver
+{
+    public function resolve(?string $templatePath): ?string
+    {
+        if ($templatePath === null) {
+            return null;
+        }
+
+        $templateContent = file_get_contents($templatePath) ?: '';
+        preg_match('/{(layout|extend) (?<layout_name>.*?)}/', $templateContent, $match);
```
MartinMystikJonas, line 27: Could there be a better way to solve this than analsis of template text by regexp? Something similar to mechanism we use in included templates? Analyse compiled template source a find nodes that rewrites layout and use default layout (passed in something like `$__layout__`)?

MartinMystikJonas:
Macros {layout} and {extends} are not presenter specific. Any template can extend another template explicitly (by using that macros). Only specific of presenters is that presenters set layout automatically without need to explicitly use macro in template. But it seems your code works only for presenter templates.

src/Resolver/LayoutResolver/LayoutPathResolver.php
lulco: Sure, with some compiler and php parser with node visitor.

I wanted to use similar way like in included templates but at that point I don’t know if it is presenter's template or component's or something else.

I wanted to apply this only to presenter's templates, didn’t know it can be used anywhere

lulco:
> Macros {layout} and {extends} are not presenter specific. Any template can extend another template explicitly (by using that macros). Only specific of presenters is that presenters set layout automatically without need to explicitly use macro in template. But it seems your code works only for presenter templates.

Yes, I didn’t know it can be used in every template. I've never used it that way.

also is little bit more complicated because you can call $this->setLayout() in presenter right?

MartinMystikJonas:
I never used setLayout but yes it can be used to change layout. But ot should be possible to handle it basically the same way as setView - we have all needed parts. We would need to collect setLayout calls, and then use it when processing presenter templates. And then pass it to template somehow (special variable seems like easiest way)

lulco:
I've added feature flag for this, so we can test it in a wild and improve later.

MartinMystikJonas (approved):

MartinMystikJonas:
Approved but we shold create issue for that so we do not forget to improve eventually

lulco:
> Approved but we shold create issue for that so we do not forget to improve eventually

Issues have been created, I'm merging this now

Issue #435, opened 2023-08-23
Can't add first class callable filter: No closure found


spaze:
When adding a filter like:
```php
$engine->addFilter('objectFilterFirstClassCallable', (new Filters())->objectFilter(...));
```
PHPStan returns "No closure found". I don't even know where to start so I'll just add a failing test (#436).

MartinMystikJonas:
This looks like limitation of roave/betterreflection which is used to analyse closure created by first class callable. It tries to analyse given closure (param and return types) by parsing known location (file+line) of closure definition (it is part of closure "object") but looks only for normal closures not for first class callables.

Honestly I have no idea how to fix this.

MartinMystikJonas:
I will check is there is any workaround for this used in PHPStan itself somewhere.

MartinMystikJonas:
Reported issue to roave/betterreflection https://github.com/Roave/BetterReflection/issues/1363

ondrejmirtes:
Hey, please let me know the place in code where this problem occurs for you.

Personally I think if you did `$scope->getType()` instead of going through BetterReflection you should be able to avoid this problem.

MartinMystikJonas:
It happens here: https://github.com/efabrica-team/phpstan-latte/blob/af52e65f0d7c5bfd29e4625ebc1f3449f0a2a660/src/Compiler/NodeVisitor/ChangeFiltersNodeVisitor.php#L239


We have Closure (real) created as first class callable which is extracted from Latte engine. And we need to know its signature to be able to preprocess code for analysis. Replace magic call by direct call to given method for which first class callable was created.

ondrejmirtes:
> we need to know its signature to be able to preprocess code for analysis

In which form do you need the signature? What about FunctionReflection/MethodReflection or ParametersAcceptor?

MartinMystikJonas:
Well, class name + method name of first class callablectarget would be enough. Or we need types of all parameters and return type to create matching placeholder - MethodReflection should be enough.

Problem is we do not have scope from plave where closure is defined - we just get closure itself from latte engine.

ondrejmirtes:
Why is this code in a node visitor and not done later where you Scope and other helpers available?

MartinMystikJonas:
We compile template by latte, use these preprocessors to make it analysable and save it. Then we run analysis of this compiled template.

ondrejmirtes:
Please show me these steps:

1) Original latte template
[...]

PR #436, opened 2023-08-23
Support first class callable, failing test

spaze:
It will fail intentionally on PHP 8.0 and lower because the syntax is supported starting with PHP 8.1, that's expected, will see later if this test can be run only on 8.1+ and how.

But it will also fail on PHP 8.2 with "No closure found" which is not expected.
https://github.com/efabrica-team/phpstan-latte/actions/runs/5946401448/job/16126895022?pr=436#step:5:28

For #435

### DIFF
--- a/.github/workflows/syntax_checker.yml
+++ b/.github/workflows/syntax_checker.yml
@@ -27,4 +27,26 @@ jobs:
           coverage: none # disable xdebug, pcov
 
       - name: Check syntax
-        run: find src tests -name "*.php" -print0 | xargs -0 -n1 -P8 php -l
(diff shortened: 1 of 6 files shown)


lulco:
@spaze nice :) you've created your first failing test case :) 

For now we don’t have any php specific test cases, but for this we will need some. Also for enums and other php 8.x specific cases

spaze:
Here's an idea how to test only things that are supported by the particular PHP version used for the test. In PHPUnit it's just a docblock comment, syntax checker in `syntax_checker.yml` may still need some tuning but at least now tests fail only to highlight the problem.

MartinMystikJonas:
Fixed by #450

PR #455, opened 2024-02-18
Add tests that should alert on `{$enum}` in templates, but passes

spaze:
Today I was changing some constants to enums, and `{$enum}` in a template fails with "Object of ... could not be converted to string" but this extension doesn't support it yet, so I thought I'd create a (failing) test at least.


MartinMystikJonas:
I finally have some time to fix issues in this project. I focused of fixing comatibility with PHPStan 2. That is now finished so I will check if this issue with enums can be solved.

MartinMystikJonas:
See #464 for fix of this issue

PR #456, opened 2024-03-22
Fixed functions handling with FunctionExecutor in new Latte and removed unformatPresenterClass of new PresenterFactory

MartinMystikJonas:
Fisrt fix: In new Latte functions are not called directly but throught FunctionExecutor and Template is always passed as first parameter on call - FunctionExecutor then handles if it is passed to function or not based on type of first parameter (it is same principle used already for filters with FilterExecutor).

Second fix: New Nette removed unformatPresenterClass from PresenterFactory - we need to reimplement it in our FakePresenterFactory used to evaluate links. If user uses own PresenterFactory it would require to add implementation of unformatPresenterClass himself.

(In the future it would be best to replace implementation used here by https://github.com/MetisFW/phpstan-nette-links - I solved these issues more universaly there)

I also fixed some minor codiing standard and PHPStan errors.

### DIFF
--- a/src/Compiler/Compiler/AbstractCompiler.php
+++ b/src/Compiler/Compiler/AbstractCompiler.php
@@ -42,6 +42,7 @@ abstract class AbstractCompiler implements CompilerInterface
         $this->strictMode = $strictMode;
         $this->filters = $filters;
         $this->functions = $functions;
+        $this->installFunctions($functions);
     }
 
     public function generateClassName(): string
@@ -93,6 +94,23 @@ abstract class AbstractCompiler implements CompilerInterface
         $providers['coreExceptionHandler'] = new CallableType($coreExceptionHandlerParameters, new VoidType());
         $phpContent .= $this->generateTypes($className . '_global', $providers);
         return $phpContent;
+    }
+
+    /**
+     * @param array<string, string|array{string, string}> $functions
+     */
+    private function installFunctions(array $functions): void
+    {
+        foreach ($functions as $name => $function) {
+            if (is_callable($function)) {
+                $this->engine->addFunction($name, $function);
+            } else {
+                // add placeholder function
+                $this->engine->addFunction($name, function (...$args) {
+                    return null;
+                });
+            }
+        }
     }
 
     abstract protected function createDefaultEngine(): Engine;
--- a/src/Compiler/Compiler/Latte2Compiler.php
+++ b/src/Compiler/Compiler/Latte2Compiler.php
@@ -50,7 +50,7 @@ final class Latte2Compiler extends AbstractCompiler
             $latteTokens = $parser->parse($templateContent);
             $className = $this->generateClassName();
             $phpContent = $compiler
-                ->setFunctions(array_keys($this->getDefaultFunctions()))
+                ->setFunctions(array_keys($this->getFunctions()))
                 ->compile(
                     $latteTokens,
                     $className,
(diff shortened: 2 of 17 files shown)


MartinMystikJonas:
I'm not sure what causes lowest dependencies tests to fail on testVariables.

How to solve composer outdated on nikic/php-parser - v5 is not compatible. Add to ignore?

lulco:
Drop support of lowest packages which not pass the test and add php parser to ignore

src/Compiler/Compiler/AbstractCompiler.php
```diff
@@ -42,6 +42,7 @@ public function __construct(?Engine $engine = null, bool $strictMode = false, ar
         $this->strictMode = $strictMode;
         $this->filters = $filters;
         $this->functions = $functions;
+        $this->installFunctions($functions);
     }
```
MartinMystikJonas, line 46: We need to install functions so Compiler handles them correctly

src/Compiler/Compiler/Latte2Compiler.php
```diff
@@ -50,7 +50,7 @@ public function compile(string $templateContent, ?string $actualClass, string $c
             $latteTokens = $parser->parse($templateContent);
             $className = $this->generateClassName();
             $phpContent = $compiler
-                ->setFunctions(array_keys($this->getDefaultFunctions()))
+                ->setFunctions(array_keys($this->getFunctions()))
```
MartinMystikJonas, line 53: This was bug in Latte 2 implementation. We did not compiled with correctly set functions.

src/Compiler/Compiler/Latte2Compiler.php
```diff
@@ -109,6 +114,25 @@ private function getEngineFiltersByReflection(): array
         return $engineFilters;
     }
 
+    /**
+     * @return array<string, callable>
+     * @throws ReflectionException
+     */
+    private function getEngineFunctionsByReflection(): array
```
MartinMystikJonas, line 121: This magix is required because Engine holds list of functions in private property. It is same approach we already use for filters

src/Compiler/NodeVisitor/AddFormClassesNodeVisitor.php
```diff
@@ -510,7 +510,7 @@ private function createClassNode(string $parentClassName, ControlHolderInterface
                 ->setReturnType('?Nette\Forms\ControlGroup');
 
             $getGroupComment = $this->createGetGroupConditionalReturnTypeComment($controlHolder->getGroups());
-            $getGroupMethod->setDocComment('/** ' . $getGroupComment . ' */');
+            $getGroupMethod->setDocComment('/** @param int|string $name' . "\n" . $getGroupComment . ' */');
```
MartinMystikJonas, line 513: Fixes PHPStan errors with latest Latte

src/Compiler/NodeVisitor/ChangeFunctionsNodeVisitor.php
```diff
@@ -284,4 +293,30 @@ private function createFunctionCallNode(string $functionName, array $args): ?Nod
             $args
         );
     }
+
+    /**
+     * @param BetterReflectionFunction|BetterReflectionMethod|ReflectionFunction $reflection
+     * @param Arg[]|VariadicPlaceholder[] $args
+     * @return Arg[]|VariadicPlaceholder[]
+     */
+    private function updateArgs($reflection, array $args): array
```
MartinMystikJonas, line 302: Remove first argument if we pass $this and function does not expect Tempalte as first argument. New latte always passes this and FunctionExecutor handles this logic

lulco (approved):

PR #458, opened 2024-09-16
Make Collector return type more explicit

staabm:
while the previous return type was correct, it did not explicitly state/protect you from someone changing the collector and make it return a empty array.

thats a major problem which can happen, and I want to make sure this error will not bite you, like it did in my projects.

for in-detail context see https://github.com/phpstan/phpstan/discussions/11701#discussioncomment-10660711

### DIFF
--- a/src/Collector/Collector/AbstractCollector.php
+++ b/src/Collector/Collector/AbstractCollector.php
@@ -35,7 +35,7 @@ abstract class AbstractCollector implements Collector
 
     /**
      * @phpstan-param array<T> $items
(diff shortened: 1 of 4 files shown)


staabm:
I did not apply the fix to all collector classes in this repo - but I think you get the point and can do the rest on your own.

if you have any questions left, feel free to ask.

I am not a phpstan-latte user myself, therefore I did not take care of making this thing perfect. My primary motivation was to make sure people don't shoot their own feet because of problems with Collectors

PR #464, opened 2025-07-23
Detect output escape parameter type errors

MartinMystikJonas:
Solves #455 

It turned out to be more general error of not validating if outputted value can be converted to string in escaping calls before output.


MartinMystikJonas:
@lulco How to solve extensions finder error? It is caused by file with Enum definition that causes it to fail. Maybe use PHP8.1+ parser in extensions finder?

lulco:
👍

lulco (approved):

PR #465, opened 2025-07-23
Support PHP 8.4

spaze:


MartinMystikJonas:
Thanks. Reported error with internal call cal be ignored by adding it to `Error\ErrorBuilder` but I am not sure what causes change of reported line in one latte file. I will investigate once it is merged

MartinMystikJonas:
I did quick check and it seems like bug in some Latte version (at least 3.0.18 which is loest PHP8.4 compatible) where comments marking latte lines in compiled PHP are generated wrong by Latte. Solution can be to mark this extension incompatible with given latte versions. Latest version of Latte works ok. Can you investigate which versions have this problem? We can then modify composer.json to exclude these versions

spaze:
Thanks for looking into this. I've already started checking the versions, will report once I have some numbers. So far I've only figured that the error with internal calls disappears in Latte 3.0.23 which is the latest version 😅 Stay tuned.

MartinMystikJonas:
It seems that this commit can be related to it because it modified how liner are marked in code generating html tags wich is exactly the code where this error occured: https://github.com/nette/latte/commit/298f4e2467d35df466a8c8f5cb39631f4c250e21

spaze:
The root cause seems to be nette/application, hmm. Requiring ^3.1.11 instead of ^3.1.6 seems to have fixed all the problems (although I thought it will fix the wrong lines problem only, but it seems to have fixed also the internal calls error, which I haven't expected).

The issue with the lines seems to be related to this commit https://github.com/nette/application/commit/88b5a83d35aef4c4917c234cc79e3e1388f7716f

I don't know (yet) what fixed the problem with the internal errors.

MartinMystikJonas:
Nice now it looks good to me. Approved @lulco ?

PR #466, opened 2025-07-25
Support PHPUnit 12

spaze:
PHPUnit 9.5 is not EOL as of now, but let's be ready when it will be
https://phpunit.de/supported-versions.html

Changes:
- test classes can't be abstract (renamed the abstract test classes to `TestCase`s)
- data providers must be static methods
- data providers are referenced by using the `DataProvider` attribute

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "nette/finder": "^2.5|^3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^9.5|^12.2",
         "nette/application": "^3.1.6",
         "nette/forms": "^3.1.12",
         "nikic/php-parser": "^5.5",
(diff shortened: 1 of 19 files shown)


spaze:
Closing, seems not worth it. Supporting both PHPUnit 9 and 12 would require some magic, excluding code coverage for 12, or using two config files or something like that. I could try harder if you think PHPUnit 12 should be supported :-)

MartinMystikJonas:
I think we should simply move to PHPUnit 12 and do not try to support both versions.

spaze:
If you want to support PHP 7.4, you'll need PHPUnit 9 unfortunately. PHPUnit 12 requires PHP >=8.3, PHPUnit 10 >=8.1.

One option is to not generate a coverage report, another is to use only modern PHP for unit testing, or create the coverage report on 8.4 for example (not sure older PHPUnit will accept the new XML config though).

PR #475, opened 2026-04-01
PHPStan reflections

MartinMystikJonas:
Last version of PHPStan breaks direct use o BetterRefection causing reflection exceptions.

I rewrited code to avoid BetterReflection completely and use only PHPStan reflections - which is correct way we should used from start anyway.

It is BC break for users using custom resolvers because signature of some methods changed. But fix should be staraighforward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,17 @@
 /composer.lock
 /.phpunit.result.cache
 /tmp
+.agents/.aiagents
+.agents/bin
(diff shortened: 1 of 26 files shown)


MartinMystikJonas:
Found it that recend update of PHPStan also broek some other things for latte analyse with generics. I will look at it.

lulco (approved):

